PR #10, opened 2021-10-06
Replace remaining uses of CTS1.

abizjak:
## Purpose

There were still some leftover uses of CTS1.


abizjak:
Thanks to @EmilLa1 for pointing it out.

EmilLa1 (approved):

PR #13, opened 2021-10-11
Use URL Json object for images in CIS1 metadata

limemloh:
## Purpose

Update CIS-1 to use URL JSON objects for images in the token metadata.
This was an oversight, and the token metadata examples already follow this.

## Changes

- Set the JSON Value type for `thumbnail` and `display` to URL JSON object instead of string.


abizjak:
Can you please start versioning and timestamping the drafts when making changes so that the evolution is clear, e.g.,  like they do for these https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04

abizjak (approved):

PR #14, opened 2021-10-18
CIS-1: Token metadata query and remove contract only error

limemloh:
## Purpose

Update CIS-1 draft to version 3

## Changes

- Add query for token metadata URL.
- Remove requirement for rejecting non-contract query calls.
- Make query calls, take a contract address for receiving the result of the query.
- Add subsection in Decisions with why there is no explicit authorization.
- Add requirement for accounts without tokens to be treated as balance is 0 for `balanceOf` and `transfer`.
- Add requirement for preserving balances to `updateOperator` and `balanceOf`.
- Add requirement for preserving operators to `balanceOf`.

## Checklist

- [x] I have performed a self-review of the changes.
- [x] I have documented my code, in particular the intent of the
      hard-to-understand areas.

### DIFF
--- a/source/CIS/cis-1.rst
+++ b/source/CIS/cis-1.rst
@@ -321,6 +321,51 @@ Requirements
 
   - A query MUST fail if the token ID is unknown with error: :ref:`INVALID_TOKEN_ID<CIS-1-rejection-errors>`.
 
+.. _CIS-1-functions-operatorOf:
+
+``operatorOf``
+^^^^^^^^^^^^^^
+
+Query operators with a list of owner address paired with an address to check for being an operator for the owner address.
+
+Parameter
+~~~~~~~~~
+
+The parameter consists of a name of the contract address and receive function to callback with the result and a list of address pairs.
+
+It is serialized as: a :ref:`CIS-1-ContractAddress` (``resultContract``) then a :ref:`CIS-1-ReceiveHookName` (``resultHook``) followed by 2 bytes for the number of queries (``n``) and then this number of queries (``queries``).
+A query is serialized as :ref:`CIS-1-Address` (``owner``) followed by :ref:`CIS-1-Address` (``address``)::
+
+  OperatorOfQuery ::= (owner: Address) (address: Address)
+
+  OperatorOfParameter ::= (resultContract: ContractAddress) (resultHook: ReceiveHookName) (n: Byte²) (queries: OperatorOfQueryⁿ)
+
+.. note::
+
+  Be aware of the size limit on contract function parameters which currently is 1024 bytes, which puts a limit on the number of queries depending on the name of the receive function.
+
+Result parameter
+~~~~~~~~~~~~~~~~
+
+The parameter for the callback receive function is a list of query and boolean pairs.
+
+It is serialized as: 2 bytes for the number of query-boolean pairs (``n``) and then this number of pairs (``results``).
+A query-boolean pair is serialized as a query (``query``) and then a byte with value 0 for false and 1 for true (``isOperator``)::
+
+  Bool ::= (0: Byte) // False
+         | (1: Byte) // True
+
+  OperatorOfQueryResult ::= (query: OperatorOfQuery) (isOperator: Bool)
+
+  OperatorOfResultParameter ::= (n: Byte²) (results: OperatorOfQueryResultⁿ)
+
+Requirements
+~~~~~~~~~~~~
+
+- The contract function MUST NOT increase or decrease the balance of any address for any token type.
+- The contract function MUST NOT add or remove any operator for any address.
+- The contract function MUST reject if any of the queries fails.
+
 .. _CIS-1-functions-tokenMetadata:
 
 ``tokenMetadata``
--- a/source/CIS/cis-1.rst
+++ b/source/CIS/cis-1.rst
@@ -464,7 +464,7 @@ A smart contract following this specification MUST reject the specified errors f
     - An address balance contains insufficient amount of tokens to complete some transfer of a token.
   * - UNAUTHORIZED
     - -42000003
-    - Sender is unauthorized to call this function. Note authorization is not mandated anywhere in this specification, but can still be introduce on top of the standard.
+    - Sender is unauthorized to call this function. Note authorization is not mandated anywhere in this specification, but can still be introduced on top of the standard.
 
 The smart contract implementing this specification MAY introduce custom error codes other than the ones specified in the table above.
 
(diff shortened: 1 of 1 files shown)


source/CIS/cis-1.rst
```diff
@@ -415,10 +464,7 @@ A smart contract following this specification MUST reject the specified errors f
     - An address balance contains insufficient amount of tokens to complete some transfer of a token.
   * - UNAUTHORIZED
     - -42000003
-    - Sender is not the address owning the tokens or an operator of the owning address. Note this can also be used if adding another authentication level on top of the standard.
-  * - CONTRACT_ONLY
-    - -42000004
-    - The sender is not a contract address.
+    - Sender is unauthorized to call this function. Note authorization is not mandated anywhere in this specification, but can still be introduce on top of the standard.
```
abizjak, line 467:
```suggestion
    - Sender is unauthorized to call this function. Note authorization is not mandated anywhere in this specification, but can still be introduced on top of the standard.
```

source/CIS/cis-1.rst
```diff
@@ -616,28 +662,33 @@ The specification only has a ``transfer`` smart contract function which takes li
 This will result in lower energy cost compared to multiple contract calls and only introduces a small overhead for single transfers.
 The reason for not also including a single transfer function is to have smaller smart contract modules, which in turn leads to saving cost on every function call.
 
+No explicit authentication
```
abizjak, line 665: Should this be authorization instead?

I suppose we also don't mandate any authentication other than that provided by the chain (so sender/owner are authenticated).

source/CIS/cis-1.rst
```diff
@@ -616,28 +662,33 @@ The specification only has a ``transfer`` smart contract function which takes li
...
+--------------------------
+
+The specification does not mandate any authentication scheme and one might expect a requirement for the owner and operators being authenticated to transfer tokens.
```
abizjak, line 668: I think authorized is the correct notion here, no?

source/CIS/cis-1.rst
```diff
@@ -273,6 +273,7 @@ Requirements
 ~~~~~~~~~~~~
 
 - The list of updates MUST be executed in order.
+- The contract function MUST NOT increase or decrease the balance of any address for any token type.
```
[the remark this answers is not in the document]
annenkov: I mean a situation like this. Let's say we had before the call
``` 
addr1 : 1
addr2 : 2 
```
After calling `updateOperator` we get

``` 
addr1 : 1
addr2 : 2
addr3 : 3 
```
So the number of tokens for the addresses in the starting state hasn't increased or decreased, but there were 3 tokens minted for `addr3`. It boils down to how the maps `Address -> TokenAmount` are treated: partial, or total with all address owning `0` by default.

source/CIS/cis-1.rst
```diff
@@ -10,14 +10,14 @@ CIS-1: Concordium Token Standard
    * - Final
      - planned Nov 1, 2021
    * - Draft version
-     - 2 (Oct 11, 2021)
+     - 3 (Oct 18, 2021)
```
abizjak, line 13: Should we also update to "final" status now?

source/CIS/cis-1.rst
```diff
@@ -257,48 +273,138 @@ Requirements
...
+  OperatorOfQueryResult ::= (query: OperatorOfQuery) (isOperator: Bool)
+
+  OperatorOfResultParameter ::= (n: Byte²) (results: OperatorOfQueryResultⁿ)
+
+Requirements
+~~~~~~~~~~~~
+
+- The contract function MUST NOT increase or decrease the balance of any address for any token type.
+- The contract function MUST NOT add or remove any operator for any address.
+- The contract function MUST reject if any of the queries fails.
```
abizjak, line 368:
```suggestion
- The contract function MUST reject if any of the queries fail.
```

source/CIS/cis-1.rst
```diff
@@ -257,48 +273,138 @@ Requirements
...
+
+The parameter consists of a name of the contract address and receive function to callback with the result and a list of token IDs.
+
+It is serialized as: a :ref:`CIS-1-ContractAddress` (``resultContract``) then a :ref:`CIS-1-ReceiveHookName` (``resultHook``) followed by 1 byte for the number of queries (``n``) and then this number of :ref:`CIS-1-TokenID` (``ids``)::
+
+  TokenMetadataParameter ::= (resultContract: ContractAddress) (resultHook: ReceiveHookName) (n: Byte²) (ids: TokenIDⁿ)
+
+.. note::
+
+  Be aware of the size limit on contract function parameters which currently is 1024 bytes, which puts a limit on the number of queries depending on the byte size of the Token ID and the name of the receive function.
```
abizjak, line 388:
```suggestion
  Be aware of the size limit on contract function parameters which is currently 1024 bytes, which puts a limit on the number of queries depending on the byte size of the Token ID and the name of the receive function.
```

source/CIS/cis-1.rst
```diff
@@ -616,28 +714,33 @@ The specification only has a ``transfer`` smart contract function which takes li
...
+The specification does not mandate any authorization scheme and one might expect a requirement for the owner and operators being authorized to transfer tokens.
+This is very much intentional and the reasoning for this is to keep the specification focused on the interface for transferring token ownership with as few restrictions as possible.
+
+Having a requirement that only owners and operators can transfer would prevent introducing any other authorization scheme on top of this specification.
+
+Adding a requirement for owners and operators being authorized to transfer tokens would prevent introducing custom contract logic rejecting transfers, such as limiting the daily transfers, temporary token lockups or non-transferrable tokens.
+
+Instead, this specification includes a requirement to ensure transfers by operators are executed as if they are sent by the owner, meaning whenever a token owner is authorized, so is an operator of the owner.
+
+Most contracts implementing this specification should probably add some authorization and not have anyone being able to transfer any token, but this is not really relevant for the interface described in this specification.
```
abizjak, line 729: Perhaps rephrase "not really relevant" to something like "not in the scope of this standard".

source/CIS/cis-1.rst
```diff
@@ -616,28 +714,33 @@ The specification only has a ``transfer`` smart contract function which takes li
...
+No token-level approval/allowance like in ERC20 and ERC721
 ----------------------------------------------------------
 
-This standard only specifies address-level operators and no authentication on per token level.
+This standard only specifies address-level operators and not token-level operators.
 The main argument is simplicity and to save energy cost on common cases, but other reasons are:
 
-- A token level authentication requires the token smart contract to track more state, which increases the overall energy cost.
-- For token smart contracts with a lot of token types, such as a smart contract with a large collection of NFTs, a token level authentication could become very expensive.
+- Token-level operators requires the token smart contract to track more state, which increases the overall energy cost.
```
abizjak, line 737:
```suggestion
- Token-level operators require the token smart contract to track more state, which increases the overall energy cost.
```

abizjak (approved):
A few minor things that would be good to resolve, but otherwise looks good.

PR #16, opened 2021-11-15
Add update proposal for protocol version 3.

abizjak:
## Purpose

Protocol update to add account aliases.

### DIFF
--- a/commentary/P3-commentary.txt
+++ b/commentary/P3-commentary.txt
@@ -2,7 +2,7 @@
 
 Protocol Version: 3
 Builds on Protocol Version: 2
-
+Protocol Name: Alpha Centauri 3.0
 
 In protocol versions 1 and 2 accounts and account addresses are in 1-1
 relationship. In protocol version 3 each account has 16777216
@@ -25,7 +25,10 @@ the same account do not change the balance of the account, apart from
 cost. Finalization, block, and baking rewards are always received on
 the account's canonical address.
 
-It is ensured that the protocol update is fully backwards compatible.
+It is ensured that the protocol update is backwards compatible such
+that all existing accounts and smart contracts are carried over
+without any changes, and all existing successful transactions retain
+their meaning.
 
 Because existing nodes do not support account aliases, node runners
 will have to update their nodes before the protocol update will take
--- a/commentary/P3-commentary.txt
+++ b/commentary/P3-commentary.txt
@@ -3,10 +3,34 @@
 Protocol Version: 3
 Builds on Protocol Version: 2
 
-This protocol update adds support for multiple addresses for the same
-account. In protocol version 3 each account has 16777216 addresses.
+
+In protocol versions 1 and 2 accounts and account addresses are in 1-1
+relationship. In protocol version 3 each account has 16777216
+addresses, namely a so-called canonical account address together with
+matching account aliases. The canonical account address is derived
+when an account is created on chain. The other 16 million addresses
+with matching initial 29 bytes are referred to as account aliases for
+the same account. Thus, accounts can be referred to by any address
+whose initial 29 bytes match.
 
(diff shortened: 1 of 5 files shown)


updates/P3.txt
```diff
@@ -0,0 +1,113 @@
...
+4. When a self-transfer is attempted for a transfer with schedule or an
+encrypted amount transfer the transaction is rejected (as before), the
+address in the rejection reason is the canonical address of the
+account that attempted the self-transfer.
+
+                               Effects
+
+1. All existing account transactions that are valid in protocol
+version 2 retain their semantics exactly; their outcomes are the same,
+as well as their effects on the state of the chain.
```
td202, line 55: This is inaccurate. Successful transactions retain their semantics. Transactions rejected in P2 may be accepted in P3, or may be rejected for the same or different reasons. These cases ought to be spelt out.

updates/P3.txt
```diff
@@ -0,0 +1,113 @@
...
+
+2. It is theoretically possible that an account creation transaction
+(credential deployment) would be deemed invalid according to protocol
+version 3. This would happen if the address of the account that would
```
td202, line 59: More accurately, it would be rejected under protocol version 3, but accepted under protocol version 1 or 2.

updates/P3.txt
abizjak: I added a clarifying sentence. I'm not going to spell out for each transaction type how the rejection reason might have changed since that would involve spelling out the rejection reason of all transactions, which does not exist at present.

commentary/P3-commentary.txt
```diff
@@ -0,0 +1,12 @@
...
+
+Protocol Version: 3
+Builds on Protocol Version: 2
+
+This protocol update adds support for multiple addresses for the same
+account. In protocol version 3 each account has 16777216 addresses.
+
+This allows each account to have aliases for different uses and
+creates a kind of sub-account structure. An account owner can give out
+different aliases for different uses to keep track of transfers and
```
concordium-cl, line 11: Did you mean "for different users"?

commentary/P3-commentary.txt
```diff
@@ -0,0 +1,12 @@
...
+assign them meaning.
```
concordium-cl, line 12: I think the commentary is a bit short. To add a bit more understanding, I've copied some snippets from other places below, as suggestions. 

- In protocol versions 1 and 2 accounts and account addresses are in 1-1
relationship. In protocol version 3 each account has 16777216 addresses, namely a so-called canonical account address together with matching account aliases. The canonical account address is derived when an account is created on chain. The other 16 million addresses with matching initial 29 bytes are referred to as account aliases for the same account. Thus, accounts can be referred to by any address whose initial 29 bytes match. 

- Each account still has one total account balance. Hence, transfers to and from aliases of an account add to and subtract from that total account balance, respectively. Transfers between different aliases of the same account do not change the balance of the account, apart from cost. Finalization, block, and baking rewards are always received on the account's canonical address.

- It is ensured that the protocol update is fully backwards compatible. 

- Because existing nodes do not support account aliases, node runners will have to update their nodes before the protocol update will take effect. The updated node software that supports the update will be released ahead of time to give node runners time to update. A new concordium-client tool will be released concurrently. Wallets do not require a new version to be compatible with the account alias functionality.

commentary/P3-commentary.txt
abizjak: No, I meant uses.

abizjak: Does it not make sense?

commentary/P3-commentary.txt
abizjak: I added this almost verbatim. Thanks.

smh1001 (approved):

concordium-cl (approved):

PR #27, opened 2022-07-14
Introduce CIS-0 Standard Detection

### LINKED ISSUE #26: Introduce standard for smart contract interface detection
We need to provide a standard for how a smart contract should expose which standards/interfaces it supports.
This should be very similar to [EIP-165](https://eips.ethereum.org/EIPS/eip-165) found on Ethereum.

Implementing this standard in a smart contract should be possible without a significant increasing in module size.

Future standard should require implementing this standard.

limemloh:
## Purpose

Introduce standard for a smart contract to publish the standard they support.
Closes #26.

## Changes

- Add new CIS-0 standard
- Add requirement for implementing CIS-0 in CIS-2.
- Add the corresponding standard identifier in the specification of CIS-1 and CIS-2.

## Checklist

- [ ] I have performed a self-review of the changes.
- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.

### DIFF
--- a/source/CIS/cis-0.rst
+++ b/source/CIS/cis-0.rst
@@ -58,7 +58,12 @@ A smart contract implementing CIS-0 MUST export the following function: :ref:`CI
 ``supports``
 ^^^^^^^^^^^^
 
-Query supported standards using a list of standard identifers.
+Query supported standards using a list of standard identifers. The response contains a corresponding result for each standard identifier, where the result is either "Standard is not supported", "Standard is supported by this contract" or "Standard is supported by using some contract address".
+
+.. note::
+
+   The result of "Standard is supported by using some contract address" can be used to add support of smart contract standards in another smart contract.
+   This could be to split up logic or to later add support for a new standard.
 
 Parameter
 ~~~~~~~~~
@@ -75,7 +80,7 @@ Response
 The function output is a list of support results, where the order of the support results matches the order of standard identifers in the parameter.
 
 It is serialized as: 2 bytes for the number (little endian) of results (``n``) and then this number of support results (``results``).
-A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by a single byte encoding the number of contract address then this number of contract addresses (``address``).
+A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by a single byte encoding the number of contract addresses, then this number of contract addresses (``address``).
 A contract address is serialized as: first 8 bytes for the index (``index``) followed by 8 bytes for the subindex (``subindex``) both little endian::
 
   ContractAddress ::= (index: Byte⁸) (subindex: Byte⁸)
--- a/source/CIS/cis-0.rst
+++ b/source/CIS/cis-0.rst
@@ -1,0 +1,117 @@
+.. _CIS-0:
+
+=========================
+CIS-0: Standard Detection
+=========================
+
+.. list-table::
+   :stub-columns: 1
+
+   * - Created
+     - Jul 14, 2022
+   * - Draft version
+     - 1
+   * - Supported versions
+     - | Smart contract version 1 or newer
+       | (Protocol version 4 or newer)
+   * - Standard identifier
+     - ``CIS-0``
+
+.. warning::
+
+   This standard is still a draft and significant changes might still be made.
+
+Abstract
+========
+
+A standardized interface for how a smart contract publish supported smart contract standards.
+Additionally, this specification includes a list of identifiers for each standard used on the Concordium blockchain.
+
+The interface provides an entrypoint for querying the smart contract, whether it supports a given standard. The smart contract respond with either: not supported, supported or supported by some contract address.
+
+Specification
+=============
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in :rfc:`2119`.
+
+.. _CIS-0-standard-identifer:
+
+Standard identifer
+-------------------
+
+Each standard MUST have a unique identifier.
+An identifier is a variable-length ASCII string up to 100 characters.
+
+The identifier is serialized as: first 1 byte for the length followed this many bytes for the ASCII encoding of the identifier::
+
+  StandardIdentifier ::= (n: Byte) (id: Byteⁿ)
+
+
+The table below contains the current standards and associated standard identifier.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 2 1 1
+
+   * - Standard
+     - Identifier
+     - Encoded
+   * - :ref:`CIS-0`
+     - ``CIS-0``
+     - ``054349532d30``
+   * - :ref:`CIS-1`
+     - ``CIS-1``
+     - ``054349532d31``
+   * - :ref:`CIS-2`
+     - ``CIS-2``
+     - ``054349532d32``
+
+.. note::
+
+   The above table will be expanded with identifers, when new standards are introduced.
+
+Contract functions
+------------------
+
+A smart contract implementing CIS-0 MUST export the following function: :ref:`CIS-0-functions-supports`.
+
+.. _CIS-0-functions-supports:
+
+``supports``
+^^^^^^^^^^^^
+
+Query supported standards using a list of standard identifers.
+
+Parameter
+~~~~~~~~~
+
+The parameter consists of a list of standard identifers.
+
+It is serialized as: 2 bytes for the number (little endian) of standard identifers and then this number of :ref:`CIS-0-standard-identifer` identifers::
+
+  SupportsParameter ::= (n : Byte²) (ids: StandardIdentifierⁿ)
+
+Response
+~~~~~~~~
+
+The function output is a list of support results, where the order of the support results matches the order of standard identifers in the parameter.
+
+It is serialized as: 2 bytes for the number (little endian) of results (``n``) and then this number of support results (``results``).
+A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by the contract address (``address``).
+A contract address is serialized as: first 8 bytes for the index (``index``) followed by 8 bytes for the subindex (``subindex``) both little endian::
+
+  ContractAddress ::= (index: Byte⁸) (subindex: Byte⁸)
+
+  SupportResult ::= (0 : Byte)                             // Standard is not supported
+                  | (1 : Byte)                             // Standard is supported by this contract
+                  | (2 : Byte) (address: ContractAddress)  // Standard is supported by using another contract address
+
+  SupportsResponse ::= (n : Byte²) (results: SupportResultⁿ)
+
+Requirements
+~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+
+   This standard is still a draft and significant changes might still be made.
+
+Abstract
+========
+
+A standardized interface for how a smart contract publish supported smart contract standards.
+Additionally, this specification includes a list of identifiers for each standard used on the Concordium blockchain.
+
+The interface provides an entrypoint for querying the smart contract, whether it supports a given standard. The smart contract respond with either: not supported, supported or supported by some contract address.
```
Bargsteen, line 30:
```suggestion
The interface provides an entrypoint for querying the smart contract about whether it supports a given standard. The smart contract responds with either: not supported, supported, or supported by some contract address.
```

Bargsteen (commented):
Looks good, but I think we need to explain what is meant by the "supported by using another contract address".
Also, we could consider changing the entrypoint name to be something that is less likely to conflict with other names. For example `supports_standards`.

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+     - ``CIS-0``
+
+.. warning::
+
+   This standard is still a draft and significant changes might still be made.
+
+Abstract
+========
+
+A standardized interface for how a smart contract publish supported smart contract standards.
```
abizjak, line 27: It is rather strange to state it as "publish" since no publishing is being done.

Perhaps "report" or "indicate"?

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in :rfc:`2119`.
+
+.. _CIS-0-standard-identifer:
+
+Standard identifer
+-------------------
+
+Each standard MUST have a unique identifier.
+An identifier is a variable-length ASCII string up to 100 characters.
```
abizjak, line 43: Why not 255 then?

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+     - Encoded
+   * - :ref:`CIS-0`
+     - ``CIS-0``
+     - ``054349532d30``
+   * - :ref:`CIS-1`
+     - ``CIS-1``
+     - ``054349532d31``
+   * - :ref:`CIS-2`
+     - ``CIS-2``
+     - ``054349532d32``
```
abizjak, line 67: I am not sure this table should be part of the standard since it means we have to keep updating the standard.

We need to have a better way of doing this.

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+
+.. note::
+
+   The above table will be expanded with identifers, when new standards are introduced.
+
+Contract functions
+------------------
+
+A smart contract implementing CIS-0 MUST export the following function: :ref:`CIS-0-functions-supports`.
```
abizjak, line 76: What is the rationale for this name. What does "functions" refer to?

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+  SupportResult ::= (0 : Byte)                             // Standard is not supported
+                  | (1 : Byte)                             // Standard is supported by this contract
+                  | (2 : Byte) (address: ContractAddress)  // Standard is supported by using another contract address
+
+  SupportsResponse ::= (n : Byte²) (results: SupportResultⁿ)
+
+Requirements
+~~~~~~~~~~~~
+
+- The number of results in the response MUST correspond to the number of the queries in the parameter.
```
abizjak, line 114: I guess you could also technically state that no duplicate results should be provided, and each query should have a response.

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,117 @@
...
+
+It is serialized as: 2 bytes for the number (little endian) of results (``n``) and then this number of support results (``results``).
+A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by the contract address (``address``).
+A contract address is serialized as: first 8 bytes for the index (``index``) followed by 8 bytes for the subindex (``subindex``) both little endian::
+
+  ContractAddress ::= (index: Byte⁸) (subindex: Byte⁸)
+
+  SupportResult ::= (0 : Byte)                             // Standard is not supported
+                  | (1 : Byte)                             // Standard is supported by this contract
+                  | (2 : Byte) (address: ContractAddress)  // Standard is supported by using another contract address
```
abizjak, line 107: Would there be a sense of allowing more than one address?

source/CIS/cis-0.rst
limemloh: no real reason I guess.
I'll make it 255

source/CIS/cis-0.rst
limemloh: I have already added the identifier in the specification of each of the current standards.
So I'll should just remove this table.

source/CIS/cis-0.rst
limemloh: I try to follow the same structure as CIS-1 and CIS-2, but I have made "functions" to singular now

source/CIS/cis-0.rst
limemloh: I have added a requirement for the contract function to reject if failing to produce a result to a query.
Which should be enough to cover these

source/CIS/cis-0.rst
```diff
@@ -97,14 +75,14 @@ Response
 The function output is a list of support results, where the order of the support results matches the order of standard identifers in the parameter.
 
 It is serialized as: 2 bytes for the number (little endian) of results (``n``) and then this number of support results (``results``).
-A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by the contract address (``address``).
+A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by a single byte encoding the number of contract address then this number of contract addresses (``address``).
```
Bargsteen, line 78:
```
number of contract addresses, then this number of contract address ...
                          ^^^
```
Added `es,`

Bargsteen (approved):
LGTM :+1:

source/CIS/cis-0.rst
Bargsteen: `functions` is still plural.

Bargsteen: And how does it match the structure of CIS-1+2? Their endpoints are called `transfer`, `balanceOf` etc.

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,95 @@
...
+
+It is serialized as: 2 bytes for the number (little endian) of results (``n``) and then this number of support results (``results``).
+A support result is serialized as either: a byte with value ``0`` for "Standard is not supported", a byte with the value ``1`` for "Standard is support by this contract" or a byte with value ``2`` for "Standard is supported by using another contract address" followed by a single byte encoding the number of contract address then this number of contract addresses (``address``).
+A contract address is serialized as: first 8 bytes for the index (``index``) followed by 8 bytes for the subindex (``subindex``) both little endian::
+
+  ContractAddress ::= (index: Byte⁸) (subindex: Byte⁸)
+
+  SupportResult ::= (0 : Byte)                             // Standard is not supported
+                  | (1 : Byte)                             // Standard is supported by this contract
+                  | (2 : Byte) (n: Byte) (address: ContractAddressⁿ)  // Standard is supported by using one of these contract addresses.
```
Bargsteen, line 85: I think we should still consider adding more details as to what this means. We could give an example with proxies.

limemloh: I have added a note in the description of the `supports` function, let me know what you think

DOBEN (commented):
Looking good. 

One difference is that Ethereum encodes the function name as well as the input parameters e.g. `bytes4(keccak256('balanceOf(address)')) == 0x12345678`. That way a smart contract A can query if smart contract B implements the function `balanceOf` as well as query that if an  `address` is sent to smart contract B that is what it expects to receive in the `balanceOf` function.

Is it possible to add support for querying if an input parameter is in the correct format (e.g. has the right length in bytes) of what the smart contract expects to receive? Maybe that is only possible if a function has specified something like the following `parameter = "BalanceOfParams"`.

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,100 @@
...
+.. _CIS-0-functions-supports:
+
+``supports``
+^^^^^^^^^^^^
+
+Query supported standards using a list of standard identifers. The response contains a corresponding result for each standard identifier, where the result is either "Standard is not supported", "Standard is supported by this contract" or "Standard is supported by using some contract address".
+
+.. note::
+
+   The result of "Standard is supported by using some contract address" can be used to add support of smart contract standards in another smart contract.
```
td202, line 65:
```suggestion
   The result of "Standard is supported by using some contract address" can be used to add support for smart contract standards in another smart contract.
```

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,100 @@
...
+-------------------
+
+Each standard MUST have a unique identifier.
+An identifier is a variable-length ASCII string up to 255 characters.
+
+The identifier is serialized as: first 1 byte for the length followed this many bytes for the ASCII encoding of the identifier::
+
+  StandardIdentifier ::= (n: Byte) (id: Byteⁿ)
+
+The table below contains the current standards and associated standard identifier.
```
td202, line 49: What table?

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,100 @@
...
+                  | (1 : Byte)                             // Standard is supported by this contract
+                  | (2 : Byte) (n: Byte) (address: ContractAddressⁿ)  // Standard is supported by using one of these contract addresses.
+
+  SupportsResponse ::= (n : Byte²) (results: SupportResultⁿ)
+
+Requirements
+~~~~~~~~~~~~
+
+- The number of results in the response MUST correspond to the number of the queries in the parameter.
+- The order of results in the response MUST correspond to the order of queries in the parameter.
```
td202, line 98:
```suggestion
- The sequence of results in the response MUST correspond to the sequence of queries in the parameter.
```

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,100 @@
...
+- The contract function MUST reject if any of the queries fail to produce a result.
```
td202, line 99: It seems that in theory it could say "not supported" for any queries that fail to produce a result. So maybe this could be MAY reject?

td202 (approved):

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,97 @@
...
+     - 1
+   * - Supported versions
+     - | Smart contract version 1 or newer
+       | (Protocol version 4 or newer)
+   * - Standard identifier
+     - ``CIS-0``
+
+.. warning::
+
+   This standard is still a draft and significant changes might still be made.
```
abizjak, line 22: Is this still relevant? When do we intend to finalize it then?

source/CIS/cis-0.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Abstract
+========
+
+A standardized interface for how a smart contract indicate supported smart contract standards.
```
abizjak, line 27: There is a typo here, it should be "indicates" or some other rewording.

abizjak (commented):
I have two comments which would be good to address.

PR #33, opened 2022-11-02
Update the limitation section in CIS2 to match P5

limemloh:
## Purpose/Change

Update section in CIS-2 specification (related to important smart contract restrictions) to match the relaxed restrictions of Protocol version 5.

### DIFF
--- a/source/CIS/cis-2.rst
+++ b/source/CIS/cis-2.rst
@@ -688,7 +688,7 @@ Smart contract limitations
 
 A number of limitations are important to be aware of:
 
-- Smart contract function parameters are limited to 65 KiB.
+- The byte size of smart contract function parameters are limited to at most 65535 B.
 - Each logged event is limited to 0.5 KiB.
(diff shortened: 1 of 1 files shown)


source/CIS/cis-2.rst
```diff
@@ -688,10 +688,8 @@ Smart contract limitations
 
 A number of limitations are important to be aware of:
 
-- Smart contract function parameters are limited to 1 KiB.
+- Smart contract function parameters are limited to 65 KiB.
```
abizjak, line 691: It is `64KiB` Or technically 65535 bytes at most. I thinkg the latter should be stated since `64KiB` is ambiguous (is it 65536 allowed, or not?)

abizjak (approved):
Please fix the limit for parameters. Otherwise good.

Bargsteen (approved):
LGTM, and I also agree with Ales ;) `65535 bytes`

PR #34, opened 2022-11-03
Fix Typos in P5

tschudid:
## Purpose

Fix typos in P5.txt

## Changes

See above.

## Checklist

- [x] I have performed a self-review of the changes.
- [x] I have documented my code, in particular the intent of the
      hard-to-understand areas.


abizjak:
The problem is that we then need **another** node build with a different hash for the protocol update. If we do other bugfixes then we should include it, otherwise I am inclined to let it be.

abizjak:
I will close this now for the above reason. If we have drastic changes then we can reopen and merge.

PR #39, opened 2023-03-07
Concordium DID Method

annenkov:
## Purpose

Specify the Concordium DID method based on Decentralized Identifiers (DIDs) v1.0 by W3C https://www.w3.org/TR/did-core/

The idea is to specify a standardized way of referring to accounts, smart contracts and public keys on Concordium, which will be later used for the Verifiable Credential Data Model.


source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+Identifier Syntax
+-----------------
+
+Concordium DID identifiers are defined by the following ABNF_:
+
+.. code-block:: ABNF
+
+  ccd-did = prefix ":" *1(network ":") (acctype / pkctype / scitype)
+  prefix  = %s"did:ccd"
+  network = "testnet" / "mainnet"
```
hamiidreza, line 54: Not clear what exactly network identifier is and it would be nice to define it somewhere (for example, for "pkctype: did-s, I have no idea what network identifier would mean.)

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+========================
+
+.. TODO add formal DID documents
+
+Account DID
+-----------
+
+The Account DID Document MUST contain the following data:
+
+- ``@context``
```
hamiidreza, line 128:
```suggestion
- ``@context`` - the attribute that expresses context information.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+- ``id`` - the DID of the account.
+- ``accountCredentials`` - a list of account holders represented by their account credentials ``credential-i`` for ``i = 1..M``, where ``M`` is the number of credentials.
+  The credentials contain several (at least one) signature verification keys ``key-i-j`` for ``j = 1..N_i``, where ``N_i`` is the number of keys, which can be different for each credential.
+  For each credential:
```
hamiidreza, line 132: Itemization is a bit unclear here (only the second item has a bullet). Also, "verificationMethod" should be a property for the DID; so, defining it as a "For each credential:" is a bit informal here

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+
+  {
+    "@context": [
+      "https://www.w3.org/ns/did/v1",
+      "Concordium DID URI"
+    ],
+    "id": "did:ccd:NET:acc:ADDR",
+    "accountCredentials": [
+      {
+        "verificationMethod": [
```
hamiidreza, line 148: Would be nice to have the formal definition of "verificationMethod" (i.e., defining its properties e.g., id, controller, type, etc)

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+Public Key Cryptography DID
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The DID document corresponding to a DID of the form
+
+``did:ccd:NET:pkc:PK``
+
+can be constructed directly from the DID without any lookup necessary.
+
+.. code-block:: json
```
hamiidreza, line 386: The code block doesn't seem very relevant. So, I would just remove it (unless it's giving some information about pkc-DID document that I cannot see?).

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+Appendices
+==========
+
+.. _concordium-did-verification-method:
+
+
+Threshold Verification Method
+-----------------------------
+
+The threshold verification method usen in Concordium DID Documents is based on a `ConditionalProof verification method <https://w3c-ccg.github.io/verifiable-conditions/>`_.
```
hamiidreza, line 431:
```suggestion
The threshold verification method used in Concordium DID Documents is based on a `ConditionalProof verification method <https://w3c-ccg.github.io/verifiable-conditions/>`_.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+This is a new type of verification method under development.
+``ConditionalProof`` features several extensions such as logical operations (``and``, ``or``), threshold and weighted threshold.
+Note that the method is not yet a W3C standard and currently has a *draft* status.
+
+The example below shows the ``2-out-of-3`` signature verifcation method.
```
hamiidreza, line 436:
```suggestion
The example below shows the ``2-out-of-3`` signature verification method.
```

hamiidreza (commented):
I added some comments/suggestions. Otherwise, looks good to me.

source/ID/concordium-did.rst
annenkov: Yes, good point, I think we should make a separate case for `pkctype`, I agree that it's not clear what network means in this context.

source/ID/concordium-did.rst
annenkov: I'll make it into a nested `vefiricationMethod`

source/ID/concordium-did.rst
annenkov: As we agreed, since the network is optional, we keep the syntax as it is, but add a note that if the network is present, it could affect the generated DID document in the future (e.g. different verification method).

source/ID/concordium-did.rst
annenkov: I added a reference to the w3c did standard

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,479 @@
...
+      "Concordium DID URI"
+    ],
+    "id": "did:ccd:sci:IND:SUBIND",
+    "owner": "did:ccd:NET:acc:ADDR"
+  }
+
+Where ``IND`` and ``SUBIND`` are the contract index and subindex.
+``NET`` and ``ADDR`` correspond to the network and to the owner's account address.
+
+Public Key Cryptography DID
```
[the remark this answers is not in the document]
annenkov: Done.

source/ID/concordium-did.rst
annenkov: Removed and added a note about the network

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A smart contract instance DID can be created by `deploying a smart contract module <deploy-module_>`_ and `initializing a smart contract instance <initialize-contract-instance_>`_ on the ``NET`` blockchain.
+The resulting DID is ``did:ccd:NET:sci:IND:SUBIND`` where ``IND``, ``SUBIND`` are the index and the subindex of the instance.
+
+Public Key Cryptography DID
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A public key cryptography DID can be created by generating a fresh Ed25519 key pair.
+The resulting DID is ``did:ccd:NET:pkc:PK`` where ``PK`` is the base58 encoded public key.
```
abizjak, line 319: I strongly recommend base16 here.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+``did:ccd:NET:sci:IND:SUBIND/EP?parameter_json=PAR[&format=FMT]``
+
+can be done by using the gRPC interface command ``InvokeInstance``.
+The entrypoint is considered a *view*: no state changes are persisted, only the result of the invocation is returned to the caller.
+The parameter ``PAR`` is passed to the entrypoint.
+
+The result of the invocation is the return value produced by the contract or an error, if the invocation failed.
+The optional query parameter ``format=FMT`` specifies the formatting of the return value.
+If not specified, the return value is in the JSON format corresponding to the embedded smart contract schema.
+If specified, the ``FMT`` value can be the following:
```
abizjak, line 388: Is it an error if this is specified and the schema is not there?

I would generally prefer not to rely on any schemas. They are optional.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+      "verificationMethod": [
+        {
+          "id": "did:ccd:NET:sci:IND:SUBIND/EP?parameter_json=PAR&format=ed25519-pk#key-0",
+          "type": "Ed25519VerificationKey2020",
+          "publicKeyMultibase": "XX"
+        }
+      ]
+    }
+
+- ``ed25519-pk-list`` - the entrypoint return value is expected to be a list of strings, e.g. ``["XX", "YY", "ZZ"]``.
```
abizjak, line 410: Why is the entrypoint not returning a list of byte arrays instead?

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+
+``did:ccd:pkc:0c7f4421e44a4385850b883e3bbf098f5a9853ef6f1a862c2ce2856381b5f5e3``
+
+A smart contract instance with the ``getKeys`` entrypoint that does not take any parameters
+
+``did:ccd:sci:321/getKeys``
+
+A smart contract instance with the ``getCredentials`` entrypoint taking a parameter
+
+``did:ccd:sci:123/getCredentials?parameter_json=QWWfHFtjPSLJdrz1beyVPzFmgS2ECbJVe1k2LXPwUGMZRjjpL``
```
abizjak, line 115: Do we really want to support JSON for this? What is the value in this as opposed to just the parameter in binary that is passed directly to the contract?

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+- ``verificationMethod`` - the account's verification method.
+  It is a nested :ref:`threshold scheme <concordium-did-verification-method>` requiring at ``T`` out of ``M`` credentials to sign; each credential uses its own threshold scheme requiring ``R_i`` out of ``N_i`` keys to sign, where ``i = 1..M``.and ``j = 1..N_i``.
+  The credentials are identified by a `DID fragment`_ ``#credential-i``, and the keys in each credentials by ``#key-j-i`` where ``i = 1..M`` and ``j = 1..N_i``.
+- ``authentication`` - authentication method for the account.
+
+The document MAY include any other public data of a Concordium account.
+
+.. note::
+
+  A `DID fragment`_ allows for referencing a particular credential, or a key in the Account DID Document.
```
abizjak, line 142: Does it mean we need to support this in the resolver? Or not?

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+              {
+                "id": "did:ccd:NET:acc:ADDR#credential-M",
+                "controller": "did:ccd:NET:acc:ADDR",
+                "type": "VerifiableCondition2021",
+                "threshold": "N",
+                "conditionThreshold": [
+                  {
+                    "id": "did:ccd:NET:acc:ADDR#key-1-M",
+                    "type": "Ed25519VerificationKey2020",
+                    "controller": "did:ccd:NET:acc:ADDR#credential-M",
```
abizjak, line 203: Why is this a controller? According to the spec this looks to be
```
An entity that has the capability to make changes to a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents). 
```

But why does credential-M have this capability? On its own it does not.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+---------------------------
+
+The goal of the Smart Contract Instance DID is to provide meta-data about the contract instance.
+At the moment, the main piece of meta-data is the Concordium account that send the initialization transaction.
+
+The Smart Contract Instance DID Document MUST contain the following data:
+
+- ``@context`` - the attribute that expresses context information.
+- ``id`` - the DID of the smart contract instance.
+- ``owner`` - a DID of an account that initialized the contract instance represented as a JSON object containing fields ``id`` and ``account``.
```
abizjak, line 236: I don't think we should include this. If we do it should say "creator" instead.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+
+Concordium DID identifiers are defined by the following ABNF_:
+
+.. code-block:: ABNF
+
+  ccd-did = prefix ":" *1(network ":") (acctype / pkctype / scitype)
+  prefix  = %s"did:ccd"
+  network = "testnet" / "mainnet"
+  acctype = "acc:" 50(base58char)
+  pkctype = "pkc:" 64(base58char)
```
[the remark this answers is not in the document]
annenkov: I think we took some of this from DID methods for other chains. We should just use what is the most meaningful for us. The same applies to other cases related to encoding.

source/ID/concordium-did.rst
annenkov: One reason to include this is that the issuers are account addresses that created an instance of a credential registry. I think there should be a way of referring to the contract owner.  We could use `creator` if this makes more sense.

source/ID/concordium-did.rst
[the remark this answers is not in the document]
annenkov: Account addresses are represented as base58, and let's use base16 for public keys.

source/ID/concordium-did.rst
annenkov: Yes, let's use base16.

source/ID/concordium-did.rst
annenkov: After discussing the formatting of return results with @abizjak , the proposed solution is the following. Instead of having the `format` parameter, we introduce the `standard` parameter, which refers to a CIS standard associated with the smart contract. For each standard, we define how the return data should be rendered. Specifically for the smart contract registries, we specify that the entrypoints that return public keys a rendered as DID documents with the appropriate  `verificationMethod`.

source/ID/concordium-did.rst
annenkov: Let's use binary data.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,546 @@
...
+
+Smart Contract Instance Reference Syntax
+----------------------------------------
+
+A DID reference is a DID with additional data, such as query parameters.
+Concordium smart contract entrypoints can be addressed using DID references.
+
+.. code-block:: ABNF
+
+  ccd-sc-ref = "/" entrypoint *1("?" param-name "=" 1*(base58char))
```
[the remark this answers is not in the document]
annenkov: @abizjak What do you think would be a better choice here: `base64` or `base16`?

source/ID/concordium-did.rst
annenkov: `credential-M` is the controller for the public key that belongs to this credential. Maybe I don't fully understand the idea. I thought that each credential (as part of an account) controls its keys.

abizjak: Well it does in the sense that the keys cannot be changed without the credential signing. But that is on its own not enough. So in that sense, if the credential is part of a multi-credential account it has to have cooperation of others to change the keys.

annenkov: Ok, then is makes sense to make the account `did:ccd:NET:acc:ADDR` to be the controller.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,613 @@
...
+
+Abstract
+=========
+
+Concordium is a layer-1 blockchain with identity built into the core protocol.
+We distinguish the following types of DIDs:
+
+- **Account DID** refer to accounts on the Concordium blockchain.
+- **Smart Contract DID** refer to smart contracts instances on the Concordium blockchain.
+- **Public Key DID** refers to a subject that knows the corresponding secret key.
```
abizjak, line 21: This is missing the IDP type.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,613 @@
...
+The Concordium DID method specification conforms to the requirements specified in `Decentralized Identifiers (DIDs) v1.0 <w3c-did-core-v1.0_>`_ [DID] published by the W3C Credentials Community Group.
+For more information about DIDs and DID method specifications, please see the `DID Primer`_ and `DID Spec`_.
+
+Abstract
+=========
+
+Concordium is a layer-1 blockchain with identity built into the core protocol.
+We distinguish the following types of DIDs:
+
+- **Account DID** refer to accounts on the Concordium blockchain.
```
abizjak, line 19: If the idea is that we use these uniformly to refer to credentials, then I think we need one to refer to the existing identity credentials. Referring to an account is not the right thing since it can have more than one credential.

In ID 2.0 setup the proof is associated with a credential registration ID.

We can add this part to the path and use the account DID, but we need to then define that. It seems like this solution with fragments
```
did:ccd:NET:acc:ADDR#credential-1
```
is worse than just being able to refer to a credential via its ID.

annenkov: It's an interesting point.
I don't know the exact use case for account DIDs. I imagine it's for off-chain verifiable credentials signed by the issuer (with the issuer's account keys). In this case, there might be various options, I believe.  It could be an account that sings according to the whole multi-sig structure, an individual credential, or a single key.
We could add a credential id to the fragment, instead of using indices like `credential-1`. But it might be more convenient to have a separate DID  type for that. Especially if the credential IDs are unique (are they?). I thought about credentials as part of the account structure, not as independent entities, but I could be wrong.

Maybe @tschudid can also add something to this discussion?

abizjak: I am thinking pragmatically. If the dapp asks for a proof that involves the user's ID credentials, issued by an identity provider, then we need to put some metadata into the proof as to what on-chain credential this is referring to. At the moment this is just a credential registration ID directly. If we want to have a uniform way of identifying, this should now be a DID. How should it look? An account can have multiple credentials from different identities.

Credential registration IDs are unique. That is enforced by the chain.

annenkov: Yes, I agree with that. My real question was whether a ruse credential can be fetched by it's ID directly, or one still neds to know the account address. If the account address is still needed, then fragments could be a way to go.
For example
`did:ccd:acc:4LxatD2pbmKe2zJWNF1QsSVRWPVgnfLnPPyfxVEor4pohZrKzc#8113641a210da36a9ffae6e6095b9cf486ca9b44fa059aa74577b0a1c0f7052d8e71168beccf299d767f3961b33aaae2`
In this case the fragment is the value of the `credId` field that I get through gRPC.

Then the account DID document will contain the appropriate fragment
```
{
    "@context": [
      "https://www.w3.org/ns/did/v1",
      "Concordium DID URI"
    ],
    "id": "did:ccd:testnet:acc:4LxatD2pbmKe2zJWNF1QsSVRWPVgnfLnPPyfxVEor4pohZrKzc",
    "verificationMethod": [
      {
        "id": "did:ccd:testnet:acc:4LxatD2pbmKe2zJWNF1QsSVRWPVgnfLnPPyfxVEor4pohZrKzc#acc-1",
        "controller": "did:ccd:NET:acc:ADDR",
        "type": "VerifiableCondition2021",
        "blockchainAccountId": "ADDR",
        "threshold": "T",
        "conditionThreshold": [
          {
            "verificationMethod": [
              {
                "id": "did:ccd:acc:4LxatD2pbmKe2zJWNF1QsSVRWPVgnfLnPPyfxVEor4pohZrKzc#8113641a210da36a9ffae6e6095b9cf486ca9b44fa059aa74577b0a1c0f7052d8e71168beccf299d767f3961b33aaae2",
                "controller": "did:ccd:NET:acc:ADDR",
                "type": "VerifiableCondition2021",
                "threshold": "R_1",
                "conditionThreshold": [
                  {
                    "id": "did:ccd:NET:acc:ADDR#key-1-1",
                    "type": "Ed25519VerificationKey2020",
                    "controller": "did:ccd:testnet:acc:4LxatD2pbmKe2zJWNF1QsSVRWPVgnfLnPPyfxVEor4pohZrKzc",
                    "publicKeyMultibase": "fXX"
                  },
                  "...",
                  {
                    "id": "did:ccd:NET:acc:ADDR#key-N_1-1",
                    "type": "Ed25519VerificationKey2020",
                    "controller": "did:ccd:testnet:acc:4LxatD2pbmKe2zJWNF1QsSVRWPVgnfLnPPyfxVEor4pohZrKzc",
                    "publicKeyMultibase": "fYY"
                  }
                ]
              }
            ]
          },
          "...",
            ]
          }
        ]
      }
    ],
 ...
  }
```

Otherwise, if it's possible to resolve a Concordium user credential directly, just by it's ID, we should introduce a new DID type.

Something like

```
did:ccd:NET:cred:8113641a210da36a9ffae6e6095b9cf486ca9b44fa059aa74577b0a1c0f7052d8e71168beccf299d767f3961b33aaae2
```

abizjak: You can look up via a credential ID directly. What you get is an account though, an account that has or has had the credential associated with it.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,741 @@
...
+
+.. note::
+  The ``publicKeyMultibase`` field contains a public key prefixed with ``f`` that denotes the base16 encoding.
+  See `The Multibase Encoding Scheme`_.
+
+
+Concordium Credential DID
+-------------------------
+
+The goal of the Account DID Document is to provide information about Concordium credentials, including a possibility to reference particular pieces of data, such as public keys.
```
abizjak, line 244: Shouldn't this talk about Credential DID document instead?

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,741 @@
...
+      "#credential-1"
+    ]
+  }
+
+
+Smart Contract Instance DID
+---------------------------
+
+The goal of the Smart Contract Instance DID is to provide meta-data about the contract instance.
+At the moment, the main piece of meta-data is the Concordium account that send the initialization transaction.
```
abizjak, line 298: Probably the more important part is the list of entrypoints.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,741 @@
...
+  `Dereferencing a DID URL`_ in the W3C Credentials Community Group draft report.
+
+
+.. code-block:: json
+
+  {
+    "@context": [
+      "https://www.w3.org/ns/did/v1",
+      "Concordium DID URI"
+    ],
```
abizjak, line 167: Are we going to keep these contexts?

annenkov: Let's drop them for now. The standard admits JSON representation that does not use contexts.

annenkov: I removed the contexts.

source/ID/concordium-did.rst
annenkov: Fixed.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+      {
+        "did:ccd:pkc:XX#key-0"
+      }
+    ]
+  }
+
+Identity Provider DID
+---------------------
+
+The goal of the Identity Provider DID is identify a Concodrium identity provider (IDP).
```
abizjak, line 351:
```suggestion
The goal of the Identity Provider DID is identify a Concordium identity provider (IDP).
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+An identity provider is an organization, approved by Concordium, that performs off-chain identification of users.
+IDPs are used in the account creation process to issue an identity.
+IDP DIDs can represent an issuer of a verifiable credential.
```
abizjak, line 354: I recommend spelling out identity providers.

I think acronyms in this case are needless

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+- ``description`` - A free form description the IDP.
+- ``verificationMethod`` - specifies a `DID verification method <did-vefication-method_>`_ for verifying a signature corresponding to the public key.
+
+.. code-block:: json
+
+  {
+    "id": "did:ccd:testnet:idp:3",
+    "name": "Digital Trust Solutions TestNet",
+    "url": "https://www.digitaltrustsolutions.nl",
+    "description": "Identity verified by Digital Trust Solutions on behalf of Concordium",
```
abizjak, line 369: We should probably use the test identity provider in this. Not DTS

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+An identity provider is an organization, approved by Concordium, that performs off-chain identification of users.
+IDPs are used in the account creation process to issue an identity.
+IDP DIDs can represent an issuer of a verifiable credential.
+
+The Identity Provider DID Document MUST contain the following data:
+- ``id`` - the DID of the IDP.
+- ``name`` - the IDP name.
+- ``url`` - A link to more information about the IDP.
+- ``description`` - A free form description the IDP.
+- ``verificationMethod`` - specifies a `DID verification method <did-vefication-method_>`_ for verifying a signature corresponding to the public key.
```
abizjak, line 361: This does't render as an itemized lits. It's all inline.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+Account DID
+^^^^^^^^^^^
+
+An account DID can be created by `opening an account <concordium-accounts_>`_ on the ``NET`` blockchain.
+The resulting DID is ``did:ccd:NET:acc:ADDR`` where ``ADDR`` is the base58 encoded account address.
+
+Concordium Credential DID
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A Concordium Credential DID is created as part of the account opening process.
```
abizjak, line 398: Or as part of adding acredential onto an existing account.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+Concordium Credential DID
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The DID document information for a DID of the form
+
+``did:ccd:NET:cred:CRED``
+
+can be resolved by looking up a credential with ID ``CRED`` on blockchain ``NET``.
+
+Data required to construct the DID document can be acquired by using the same gRPC interface command ``GetAccountInfo`` as for Concodrium account DIDs.
```
abizjak, line 452:
```suggestion
Data required to construct the DID document can be acquired by using the same gRPC interface command ``GetAccountInfo`` as for Concordium account DIDs.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+
+Update
+------
+
+At this time Concordium does not support the update of DID documents.
+
+Deactivate
+----------
+
+At this time Concordium does not support deactivation of DID documents.
```
abizjak, line 577: Credentials can be removed. Not sure if this matters for this or not.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+
+.. code-block:: console
+
+    $concordium-client raw GetIdentityProviders
+
+
+Update
+------
+
+At this time Concordium does not support the update of DID documents.
```
abizjak, line 572: Credential and account keys can be updated, not sure if that matters.

abizjak (approved):
Looks good overall, I have some minor comments/concerns.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+
+Smart Contract Instance Reference Syntax
+----------------------------------------
+
+A DID reference is a DID with additional data, such as a path and query parameters.
+Concordium smart contract entrypoints can be addressed using DID references.
+
+.. code-block:: ABNF
+
+  ccd-sc-ref = "/" entrypoint *1("/" parameter) *1("?" "standard" "=" standard)
```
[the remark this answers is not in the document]
annenkov: Yes, it should be removed, I'll do that.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+  parameter = base16char
+  entrypoint = 1*(ALPHA / DIGIT / punctuation)
+  standard = "CIS" "-" 1*(DIGIT)
```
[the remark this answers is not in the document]
annenkov: Will be removed.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+  `Dereferencing a DID URL`_ in the W3C Credentials Community Group draft report.
+
+
+.. code-block:: json
+
+  {
+    "id": "did:ccd:NET:acc:ADDR",
+    "verificationMethod": [
+      {
+        "id": "did:ccd:NET:acc:ADDR#acc-1",
```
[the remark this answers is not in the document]
annenkov: The fragment `#acc-1` is used to identify a verification method for the account. It could be any string; `acc-1` is confusing, so I'll change it to, e.g. `acc-vm`.  It is used as a relative reference in `authentication`. That is, this verification method is used to authenticate operations for the account.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+Threshold Verification Method
+-----------------------------
+
+The threshold verification method used in Concordium DID Documents is based on a `ConditionalProof verification method <https://w3c-ccg.github.io/verifiable-conditions/>`_.
+This is a new type of verification method under development.
+``ConditionalProof`` features several extensions such as logical operations (``and``, ``or``), threshold and weighted threshold.
+Note that the method is not yet a W3C standard and currently has a *draft* status.
+
+The example below shows the ``2-out-of-3`` signature verification method.
+It uses the ``ConditionalProof2022`` verification method.
```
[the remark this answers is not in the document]
annenkov: `2022` is a version of the `ConditionalProof` verification method, apparently. This is how it is described in this document https://w3c-ccg.github.io/verifiable-conditions/#the-conditionalproof-type

source/ID/concordium-did.rst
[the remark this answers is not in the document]
annenkov: Fiexd here and in the credential DID document as well.

source/ID/concordium-did.rst
annenkov: Fixed.

source/ID/concordium-did.rst
annenkov: Fixed.

source/ID/concordium-did.rst
annenkov: Fixed.

source/ID/concordium-did.rst
annenkov: Fixed.

source/ID/concordium-did.rst
annenkov: I added some general info referencing Rust code documentation. Here and below for deactivation.

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,661 @@
...
+
+Deactivate
+----------
+
+At this time Concordium does not support deactivation of DID documents.
+
+Security Considerations
+=======================
+
+The ``did:ccd`` method is built on top the Concordium blockchain, a public permissionless DLT.
```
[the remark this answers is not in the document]
tschudid: DL makes more sense

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+Concordium Credential DID
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A Concordium Credential DID can be deactivated by removing the corresponding credential with an `update credentials transaction <https://docs.rs/concordium_base/1.2.0/concordium_base/transactions/construct/fn.update_credentials.html>`_.
+
+Security Considerations
+=======================
+
+The ``did:ccd`` method is built on top the Concordium blockchain, a public permissionless DLT.
```
tschudid, line 597:
```suggestion
The ``did:ccd`` method is built on top the Concordium blockchain, a public permissionless DL.
```

tschudid (approved):
LGTM

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+-----------
+
+The goal of the Account DID Document is to provide information about the account authentication data, including a possibility to reference particular pieces of data, such as public keys.
+In order to do that, it specifies a `DID verification method <did-vefication-method_>`_ that reflects the account authentication data: public keys grouped into credentials.
+
+The Account DID Document MUST contain the following data:
+
+- ``id`` - the DID of the account.
+- ``verificationMethod`` - the account's verification method.
+  It is a nested :ref:`threshold scheme <concordium-did-verification-method>` requiring at ``T`` out of ``M`` credentials to sign; each credential uses its own threshold scheme requiring ``R_i`` out of ``N_i`` keys to sign, where ``i = 1..M``.and ``j = 1..N_i``.
```
dg-concordium, line 139: The "at" before T is confusing. Was that meant to be something else?

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+    "authentication": [
+      "#credential-vm"
+    ]
+  }
+
+
+Smart Contract Instance DID
+---------------------------
+
+The goal of the Smart Contract Instance DID is to provide meta-data about the contract instance.
```
dg-concordium, line 283:
```suggestion
The goal of the Smart Contract Instance DID is to provide metadata about the contract instance.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+        "name": "issuerKeys"
+      },
+      { "id": "did:ccd:sci:IND:SUBIND#entrypoint-revocationKey",
+        "name": "revocationKey"
+      }
+    ]
+  }
+
+Where ``IND`` and ``SUBIND`` are the contract index and subindex.
+``NET`` and ``ADDR`` correspond to the network and to the owner's account address.
```
dg-concordium, line 313:
```suggestion
``NET`` and ``ADDR`` correspond to the network and to the owner's account address respectively.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+        "did:ccd:pkc:XX#key-0"
+      }
+    ]
+  }
+
+Identity Provider DID
+---------------------
+
+The goal of the Identity Provider DID is identify a Concordium identity provider.
+An identity provider is an organization, approved by Concordium, that performs off-chain identification of users.
```
dg-concordium, line 352:
```suggestion
An identity provider (IDP) is an organization, approved by Concordium, that performs off-chain identification of users.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+``did:ccd:NET:acc:ADDR``
+
+can be resolved by looking up an account with address ``ADDR`` on blockchain ``NET``.
+
+Data required to construct the DID document can be acquired by using the gRPC interface command ``GetAccountInfo``.
+
+See the details in the `gRPC v2 documentation`_.
+
+From the command line, ``concordium-client`` allows to retrieve the data in the following way:
```
dg-concordium, line 435:
```suggestion
From the command line, ``concordium-client`` allows retrieving the data in the following way:
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+``did:ccd:NET:sci:IND:SUBIND``
+
+can be resolved by looking up a smart contract instance with indices ``IND``, ``SUBIND`` on blockchain ``NET``.
+This includes a lookup of the owner's account.
+
+Data required to construct the DID document can be acquired by using the gRPC interface command ``GetInstanceInfo``.
+
+See the details in the `gRPC v2 documentation`_.
+
+From the command line, ``concordium-client`` allows for retrieving the data in the following way:
```
dg-concordium, line 472:
```suggestion
From the command line, ``concordium-client`` allows retrieving the data in the following way:
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+Dereferencing a DID reference of the form
+
+``did:ccd:NET:sci:IND:SUBIND/EP[/PAR]``
+
+can be done by using the gRPC interface command ``InvokeInstance``.
+The entrypoint is considered a *view*: no state changes are persisted, only the result of the invocation is returned to the caller.
+The parameter ``PAR`` is passed to the entrypoint.
+
+The result of the invocation is the return value produced by the contract or an error, if the invocation failed.
```
dg-concordium, line 496:
```suggestion
The result of the invocation is the return value produced by the contract or an error if the invocation fails.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+.. code-block:: json
+
+  {
+    "type" : "raw",
+    "data": "BASE16DATA"
+  }
+
+Where ``BASE16DATA`` is a base16-encoded return value.
+
+From the command line, ``concordium-client`` allows for invoking a smart contract instance in the following way:
```
dg-concordium, line 520:
```suggestion
From the command line, ``concordium-client`` allows invoking a smart contract instance in the following way:
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+``did:ccd:NET:idp:INDEX``
+
+can be resolved by looking up an identity provider ``INDEX`` on blockchain ``NET``.
+
+Data required to construct the DID document can be acquired by using the gRPC interface command ``GetIdentityProviders``.
+
+See the details in the `gRPC v2 documentation`_.
+
+From the command line, ``concordium-client`` allows to retrieve the data in the following way:
```
dg-concordium, line 563:
```suggestion
From the command line, ``concordium-client`` allows retrieving the data in the following way:
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+
+Update
+------
+
+Account DID
+^^^^^^^^^^^
+
+It is possible to update Account DID documents by sending an `update credentials transaction <https://docs.rs/concordium_base/1.2.0/concordium_base/transactions/construct/fn.update_credentials.html>`_.
+This type of transactions allows for adding/removing credentials and changing the signature threshold.
```
dg-concordium, line 577:
```suggestion
This type of transaction allows for adding/removing credentials and changing the signature threshold.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+Concordium Credential DID
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is possible to update Concordium Credential DID documents by sending an `update credentials keys transaction <https://docs.rs/concordium_base/1.2.0/concordium_base/transactions/send/fn.update_credential_keys.html>`_.
+This type of transactions allows for updating the keys associated with a Concordium credential and the corresponding signature threshold.
```
dg-concordium, line 583:
```suggestion
This type of transaction allows for updating the keys associated with a Concordium credential and the corresponding signature threshold.
```

source/ID/concordium-did.rst
```diff
@@ -0,0 +1,676 @@
...
+
+The ``did:ccd`` method is built on top the Concordium blockchain, a public permissionless distributed ledger (DL).
+Security of the DID method reduces to the security of the underlying blockchain protocol.
+This concerns attacks such as eavesdropping, replay, message insertion, deletion, modification, denial of service, amplification, and man-in-the-middle.
+
+Parties SHOULD run a full node of the underlying blockchain protocol to ensure that they can read and write securely to the DL.
+
+Authorization is performed by means of digital signature keys.
+Leakage of private keys allows an attacker to take control.
+Parties therefore MUST handle private keys with care.
```
dg-concordium, line 605:
```suggestion
Therefore, parties MUST handle private keys with care.
```

dg-concordium (approved):
Just a few minor language, grammar corrections.

source/ID/concordium-did.rst
annenkov: Thanks for pointing this out; "at" should be removed :)

source/ID/concordium-did.rst
annenkov: @abizjak suggested [here](https://github.com/Concordium/concordium-update-proposals/pull/39/files/35062a110099c3074be5c0151507facf54bc6c83#r1222690694) not use the acronym.
I guess we could use "IDP" if it's recognizable, that is, used in other parts of our documentation.

dg-concordium: It is written out everywhere in the documentation. But the general rule is that as long as you write it out the first time and follow it in parentheses with the acronym, then it's fine to use the acronym in the rest of the document. But I will leave the final decision to you and Ales.

dg-concordium: That was why I put in IDP next to this. Because IDP is used elsewhere in the same doc.

annenkov: > That was why I put in IDP next to this. Because IDP is used elsewhere in the same doc.

@dg-concordium you're right, I thought I expanded the acronym in all places in the document, but it's used in several places, my bad. So, either I spell it out in full everywhere or introduce an acronym.

annenkov: @dg-concordium I'll add the acronym in parentheses, as you suggested. It's used many times, so it makes sense to have it.

source/ID/concordium-did.rst
annenkov: Done.

PR #45, opened 2023-05-01
CIS-4: Credential Registry Standard

annenkov:
## Purpose

A standard for a registry smart contract for verifiable credentials.

### DIFF
--- a/source/CIS/cis-4.rst
+++ b/source/CIS/cis-4.rst
@@ -28,6 +28,12 @@ The interface provides functionality for the following roles of users:
 - *verifiers* can query credential status and data that are used to check a verifiable presentation requested from a holder;
 - *revocation authorities* can revoke credentials by signing a revocation message.
 
+Each contract instance MUST store public data of VCs of the same type.
+A *credential type* is a string that corresponds to the name of the credential schema, which credentials in the registry are based on.
+A `VCs JSON schema <https://w3c.github.io/vc-json-schema/>`_ is a JSON document containing the `JSON schema <http://json-schema.org/>`_ describing the attributes of a VC and some metadata about the schema.
+
+.. TODO: refer to the Concordium VC Data Model documentation, once we have it.
+
 Specification
 =============
 
@@ -197,13 +203,13 @@ It is serialized as: First byte encodes the length (``n``) of the credential typ
 
 Basic data for a verifiable credential.
 
-It is serialized as a credential holder identifier :ref:`CIS-4-PublicKeyEd25519` (``holder_id``), a flag whether the credential can be revoked by the holder :ref:`CIS-4-Bool` (``holder_revocable``), a :ref:`CIS-4-Timestamp` from which the credential is valid (``valid_from``), an optional :ref:`CIS-4-Timestamp` until which the credential is valid (``valid_until``), and the credential type :ref:`CIS-4-CredentialType` (``credential_type``).
+It is serialized as a credential holder identifier :ref:`CIS-4-PublicKeyEd25519` (``holder_id``), a flag whether the credential can be revoked by the holder :ref:`CIS-4-Bool` (``holder_revocable``), a :ref:`CIS-4-Timestamp` from which the credential is valid (``valid_from``), an optional :ref:`CIS-4-Timestamp` until which the credential is valid (``valid_until``), and a reference to the credential metadata :ref:`CIS-4-MetadataUrl` (``metadata_url``).
 The optional timestamp is serialized as 1 byte to indicate whether a timestamp is included, if its value is 0, then no timestamp is present, if the value is 1 then the :ref:`CIS-4-Timestamp` bytes follow::
 
   OptionalTimestamp ::= (0: Byte)
                       | (1: Byte) (timestamp: Timestamp)
   CredentialInfo ::= (holder_id: CredentialHolderId) (holder_revocable: Bool) (valid_from: Timestamp)
-                     (valid_until: OptionTimestamp) (credential_type: CredentialType) (metadata_url: MetadataUrl)
+                     (valid_until: OptionTimestamp) (metadata_url: MetadataUrl)
 
 .. note::
   The timestamp ``valid_until`` is optional; if it is not included (indicated by the 0 tag), then the credential never expires.
@@ -235,7 +241,7 @@ A smart contract implementing this standard MUST export the following functions:
 - :ref:`CIS-4-functions-credentialEntry`
 - :ref:`CIS-4-functions-credentialStatus`
 - :ref:`CIS-4-functions-issuer`
-- :ref:`CIS-4-functions-issuerMetadata`
+- :ref:`CIS-4-functions-registryMetadata`
 - :ref:`CIS-4-functions-registerCredential`
 - :ref:`CIS-4-functions-revokeCredentialIssuer`
 - :ref:`CIS-4-functions-revokeCredentialHolder`
@@ -318,19 +324,21 @@ Response
 The function output is the issuer's public key.
 It is serialized as :ref:`CIS-4-PublicKeyEd25519`.
 
-.. _CIS-4-functions-issuerMetadata:
+.. _CIS-4-functions-registryMetadata:
 
-``issuerMetadata``
-^^^^^^^^^^^^^^^^^^
+``registryMetadata``
+^^^^^^^^^^^^^^^^^^^^
 
-Query the issuer's metadata URL.
+Query the registry's metadata.
 
 Response
 ~~~~~~~~
 
-The function output is the issuer's metadata URL.
+The function output is the issuer's metadata URL, the credential type and schema for the credentials stored in the registry.
 
-It is serialized as :ref:`CIS-2-MetadataUrl`.
+It is serialized as the issuer's :ref:`CIS-4-MetadataUrl` (``issuer_metadata``) followed by the credential type of the registry :ref:`CIS-4-CredentialType` (``credential_type``) and the corresponding credential JSON schema reference :ref:`CIS-4-SchemaRef` (``credential_schema``)::
+
+  MetadataResponse := (issuer_metadata: MetadataUrl) (credential_type: CredentialType) (credential_schema: SchemaRef)
 
 .. _CIS-4-functions-registerCredential:
 
@@ -706,13 +714,18 @@ The credential metadata is stored off-chain and MUST be a JSON (:rfc:`8259`) fil
     - The name to display for the credential.
   * - ``logo``
     - URL JSON object
-    - An image URL for displaying the credential.
+    - An image URL for displaying the credential. The RECOMMENDED size on the image is 40x40.
   * - ``background_color``
     - string
     - A hex code of the background color for displaying the credential.
   * - ``image`` (optional)
     - URL JSON object
-    - A background image URL for displaying the credential.
+    - A background image URL for displaying the credential. The RECOMMENDED size on the image is 327x120.
+  * - ``localization`` (optional)
+    - JSON object with locales as field names (:rfc:`5646`) and field values are URL JSON objects linking to JSON files.
+    - URLs to JSON files with localized token metadata.
+
+.. TODO: check the actual image sizes before finalizing the standard.
 
 Where URL JSON object the same as in :ref:`CIS-4-events-IssuerMetadata`.
 
@@ -734,7 +747,27 @@ Example credential metadata
     "image": {
       "url": "https://concordium.com/employment/vc-background.png",
     }
+    "localization": {
+      "da-DK": {
+        "url": "https://location.of/the/danish/metadata.json",
+        "hash": "624a1a7e51f7a87effbf8261426cb7d436cf597be327ebbf113e62cb7814a34b"
+      }
+    }
   }
+
+Note that that URL addresses for images can come with or without the ``hash`` attribute.
+In the example, the ``image`` attribute value does not specify ``hash``.
+In this case, no content integrity check will be performed.
+
+The danish localization JSON file could be:
+
+.. code-block:: json
+
+  {
+    "degree": "Grad",
+    "graduationDate": "Dimissionsdato",
+  }
+
 
 .. _CIS-4-smart-contract-limitations:
 
--- a/source/CIS/cis-4.rst
+++ b/source/CIS/cis-4.rst
@@ -442,7 +442,7 @@ Requirements
     - The signature was intended for a different entrypoint.
     - The signature is expired.
     - The signature cannot be validated.
-      The smart contract logic SHOULD practice its best efforts to ensure that only the holder can generate and authorize a revocation request with a valid signature.
+      The smart contract logic SHOULD use all possible efforts to ensure that only the holder can generate and authorize a revocation request with a valid signature.
 
 .. _CIS-4-functions-revokeCredentialOther:
 
@@ -484,7 +484,7 @@ Requirements
     - The signature was intended for a different entrypoint.
     - The signature is expired.
     - The signature can not be validated.
-      The smart contract logic SHOULD practice its best efforts to ensure that only the revocation authority can generate and authorize a revocation request with a valid signature.
+      The smart contract logic SHOULD use all possible efforts to ensure that only the revocation authority can generate and authorize a revocation request with a valid signature.
 
 .. _CIS-4-functions-registerRevocationKeys:
 
--- a/source/CIS/cis-4.rst
+++ b/source/CIS/cis-4.rst
@@ -545,7 +545,7 @@ The event records the credential identifier, the credential type, and the corres
 The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 255, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
 
   CredentialEventData ::= (credential_id: CredentialHolderId) (schema_ref: SchemaRef) (credential_type: CredentialType)
-  RegisterCredentialEvent ::= (255: Byte) (data: CredentialEventData)
+  RegisterCredentialEvent ::= (249: Byte) (data: CredentialEventData)
 
 ``RevokeCredentialEvent``
 ^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -558,7 +558,7 @@ The ``RevokeCredentialEvent`` event is serialized as: first a byte with the valu
   Revoker ::= (0: Byte)                         // Issuer
             | (1: Byte)                         // Holder
             | (2: Byte) (key: PublicKeyEd25519) // Other
-  RevokeCredentialEvent ::= (254: Byte) (credential_id: CredentialHolderId) (revoker: Revoker) (reason: OptionReason)
+  RevokeCredentialEvent ::= (248: Byte) (credential_id: CredentialHolderId) (revoker: Revoker) (reason: OptionReason)
 
 ``IssuerMetadata``
 ^^^^^^^^^^^^^^^^^^
@@ -569,7 +569,7 @@ It consists of a URL for the location of the metadata for the issuer with an opt
 
 The ``IssuerMetadata`` event is serialized as: first a byte with the value of 253, followed by :ref:`CIS-2-MetadataUrl` (``metadata``)::
 
-  IssuerMetadata ::= (253: Byte) (metadata: MetadataUrl)
+  IssuerMetadata ::= (247: Byte) (metadata: MetadataUrl)
 
 ``CredentialMetadataEvent``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -577,9 +577,9 @@ The ``IssuerMetadata`` event is serialized as: first a byte with the value of 25
 A ``CredentialMetadataEvent`` event MUST be logged when updating the credential metadata.
 It consist of a credential ID and a URL for the location of the metadata for this credential with an optional SHA256 checksum of the content.
 
-The ``TokenMetadataEvent`` event is serialized as: first a byte with the value of 252, followed by the token ID :ref:`CIS-2-TokenID` (``id``), and then a :ref:`CIS-2-MetadataUrl` (``metadata``)::
+The ``CredentialMetadataEvent`` event is serialized as: first a byte with the value of 252, followed by the token ID :ref:`CIS-2-TokenID` (``id``), and then a :ref:`CIS-2-MetadataUrl` (``metadata``)::
 
-  TokenMetadataEvent ::= (252: Byte) (id: CredentialHolderId) (metadata: MetadataUrl)
+  CredentialMetadataEvent ::= (246: Byte) (id: CredentialHolderId) (metadata: MetadataUrl)
 
 
 ``CredentialSchemaRefEvent``
@@ -590,7 +590,7 @@ It consist of a credential type and a URL for the location of the schema for thi
 
 The ``CredentialSchemaRefEvent`` event is serialized as: first a byte with the value of 251, followed by the token ID :ref:`CIS-4-CredentialType` (``type``), and then a :ref:`CIS-4-SchemaRef` (``schema_ref``)::
 
-  CredentialSchemaRefEvent ::= (251: Byte) (type: CredentialType) (schema_ref: SchemaRef)
+  CredentialSchemaRefEvent ::= (245: Byte) (type: CredentialType) (schema_ref: SchemaRef)
 
 ``RevocationKeyEvent``
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -602,7 +602,7 @@ The ``RevocationKeyEvent`` event is serialized as: first a byte with the value o
 
   RevocationKeyAction ::= (0: Byte)    // Register
                         | (1: Byte)    // Remove
-  RevocationKeyEvent ::= (250: Byte) (action: RevocationKeyAction)
+  RevocationKeyEvent ::= (244: Byte) (action: RevocationKeyAction)
 
 Issuer metadata JSON
 --------------------
--- a/source/CIS/cis-4.rst
+++ b/source/CIS/cis-4.rst
@@ -75,9 +75,9 @@ It is serialized as :ref:`CIS-4-PublicKeyEd25519`.
 
 A URL and optional checksum for metadata stored outside of this contract.
 
-It is serialized as: 2 bytes for the length of the metadata url (``n``) and then this many bytes for the url to the metadata (``url``) followed by an optional checksum.
+It is serialized as: 2 bytes for the length of the metadata URL (``n``) and then this many bytes for the URL to the metadata (``url``) followed by an optional checksum.
 The checksum is serialized by 1 byte to indicate whether a hash of the metadata is included.
-If its value is 0, then there is no hash, if the value is 1 then 32 bytes for a SHA256 hash (``hash``) follows::
+If its value is 0, then there is no hash; if the value is 1, then 32 bytes for a SHA256 hash (``hash``) follows::
 
   MetadataChecksum ::= (0: Byte)
                      | (1: Byte) (hash: Byte³²)
@@ -90,9 +90,9 @@ If its value is 0, then there is no hash, if the value is 1 then 32 bytes for a 
 ^^^^^^^^^^^^^^^^^^^
 
 An address of a contract instance.
-It consists of an index and a subindex both unsigned 64-bit integers.
+It consists of an index and a subindex, both unsigned 64-bit integers.
 
-It is serialized as: First 8 bytes for the index (``index``) followed by 8 bytes for the subindex (``subindex``) both little endian::
+It is serialized as: First 8 bytes for the index (``index``) followed by 8 bytes for the subindex (``subindex``), both little-endian::
 
   ContractAddress ::= (index: Byte⁸) (subindex: Byte⁸)
 
@@ -124,10 +124,10 @@ It is serialized as: First 2 bytes encode the length (``n``) of the entrypoint n
 ``Timestamp``
 ^^^^^^^^^^^^^
 
-A timestamp given in milliseconds since unix epoch.
+A timestamp given in milliseconds since Unix epoch.
 It consists of an unsigned 64-bit integer.
 
-It is serialized as 8 bytes in little endian::
+It is serialized as 8 bytes in little-endian::
 
   Timestamp ::= (milliseconds: Byte⁸)
 
@@ -138,7 +138,7 @@ It is serialized as 8 bytes in little endian::
 
 An unsigned 64-bit integer number that increases sequentially to protect against replay attacks.
 
-It is serialized as 8 bytes in little endian::
+It is serialized as 8 bytes in little-endian::
 
   Nonce ::= (nonce: Byte⁸)
 
@@ -169,7 +169,7 @@ It is serialized as 64 bytes::
 ``SigningData``
 ^^^^^^^^^^^^^^^
 
-Signing data contains a metadata for the signature that is used to check whether the signed message is designated for the right contract and entrypoint, and it is not expired.
+Signing data contains metadata for the signature that is used to check whether the signed message is designated for the correct contract and entrypoint, and that it is not expired.
 
 It is serialized as :ref:`CIS-4-ContractAddress` (``contract_address``), :ref:`CIS-4-EntrypointName` (``entrypoint``), :ref:`CIS-4-Nonce` (``nonce``), and :ref:`CIS-4-Timestamp` (``timestamp``)::
 
@@ -190,7 +190,7 @@ Serialized in the same way as :ref:`CIS-4-MetadataUrl`.
 ``CredentialType``
 ^^^^^^^^^^^^^^^^^^
 
-A short string (up to 256 characters) in the UTF-8 encoding.
+A short string (up to 256 characters) in UTF-8 encoding.
 The string describes the credential type that is used to identify which schema the credential is based on.
 It corresponds to a value of the ``name`` attribute of the credential schema.
 
@@ -206,7 +206,7 @@ It is serialized as: First byte encodes the length (``n``) of the credential typ
 Basic data for a verifiable credential.
 
 It is serialized as a credential holder identifier :ref:`CIS-4-PublicKeyEd25519` (``holder_id``), a flag whether the credential can be revoked by the holder :ref:`CIS-4-Bool` (``holder_revocable``), a :ref:`CIS-4-Timestamp` from which the credential is valid (``valid_from``), an optional :ref:`CIS-4-Timestamp` until which the credential is valid (``valid_until``), and a reference to the credential metadata :ref:`CIS-4-MetadataUrl` (``metadata_url``).
-The optional timestamp is serialized as 1 byte to indicate whether a timestamp is included, if its value is 0, then no timestamp is present, if the value is 1 then the :ref:`CIS-4-Timestamp` bytes follow::
+The optional timestamp is serialized as 1 byte to indicate whether a timestamp is included. If its value is 0, then no timestamp is present; if the value is 1, then the :ref:`CIS-4-Timestamp` bytes follow::
 
   OptionalTimestamp ::= (0: Byte)
                       | (1: Byte) (timestamp: Timestamp)
@@ -231,7 +231,7 @@ It is serialized as 1 byte where ``0`` correponds to the status ``Active``, ``1`
                      | (2: Byte) // Expired
                      | (3: Byte) // NotActivated
 
-See requirements for :ref:`CIS-4-functions-credentialStatus` for details of how statues are returned.
+See requirements for :ref:`CIS-4-functions-credentialStatus` for details of how statuses are returned.
 
 .. _CIS-4-functions:
 
@@ -353,14 +353,14 @@ Register public data for a new credential.
   This standard does not specify how the issuer is authenticated.
   Implementations can use various mechanisms.
   For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
-  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
 
 Parameter
 ~~~~~~~~~
 
 The parameter is credential information that is used to create an entry in the registry.
 
-It is serialized as :ref:`CIS-4-CredentialInfo` (``credential_info``) followed by auxiliary data, which is serialized as 2 bytes to encode the length (``n``) of the vector of keys, followed by this many bytes of data::
+It is serialized as :ref:`CIS-4-CredentialInfo` (``credential_info``), followed by auxiliary data, which is serialized as 2 bytes to encode the length (``n``) of the vector of keys, followed by this many bytes of data::
 
   AuxData ::= (n: Byte²) (data: Byteⁿ)
   RegisterCredentialParameter ::= (credential_info: CredentialInfo) (auxiliary_data: AuxData)
@@ -382,7 +382,7 @@ Revoke a credential by the issuer's request.
   This standard does not specify how the issuer is authenticated.
   Implementations can use various mechanisms.
   For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
-  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
 
 Parameter
 ~~~~~~~~~
@@ -390,7 +390,7 @@ Parameter
 The parameter is the credential ID :ref:`CIS-4-CredentialHolderId` and an optional string in the UTF-8 encoding that indicates the revocation reason.
 
 It is serialized as :ref:`CIS-4-CredentialHolderId` followed by 1 byte to indicate whether a reason is included.
-If its value is 0, then no reason string is present, if the value is 1 then the bytes corresponding to the reason string follow.
+If its value is 0, then no reason string is present; if the value is 1, then the bytes corresponding to the reason string follow.
 The optional revocation reason is followed by auxiliary data, which is serialized as 2 bytes to encode the length (``n``) of the data vector, followed by this many bytes of data::
 
   OptionalReason ::= (0: Byte)
@@ -497,7 +497,7 @@ Register public keys that can be used by revocation authorities.
   This standard does not specify how the issuer is authenticated.
   Implementations can use various mechanisms.
   For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
-  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
 
 Parameter
 ~~~~~~~~~
@@ -527,7 +527,7 @@ Make a list of public keys unavailable to revocation authorities.
   This standard does not specify how the issuer is authenticated.
   Implementations can use various mechanisms.
   For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
-  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
 
 Parameter
 ~~~~~~~~~
@@ -576,7 +576,7 @@ A custom event SHOULD NOT have a first byte colliding with any of the events def
 A ``RegisterCredentialEvent`` event MUST be logged when a new credential is registered.
 The event records the credential identifier, the credential type, and the corresponding schema reference.
 
-The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 249, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
+The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 249, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), and a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
 
   CredentialEventData ::= (credential_id: CredentialHolderId) (schema_ref: SchemaRef) (credential_type: CredentialType)
   RegisterCredentialEvent ::= (249: Byte) (data: CredentialEventData)
@@ -585,9 +585,9 @@ The ``RegisterCredentialEvent`` event is serialized as: first a byte with the va
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 A ``RevokeCredentialEvent`` event MUST be logged when a credential is revoked.
-The event records the credential identifier, who requested the revocation (the holder, the issuer or a revocation authority), and an optional string with a short comment on the revocation reason.
+The event records the credential identifier who requested the revocation (the holder, the issuer, or a revocation authority), and an optional string with a short comment on the revocation reason.
 
-The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 248, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`; ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
+The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 248, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`. ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
 
   Revoker ::= (0: Byte)                         // Issuer
             | (1: Byte)                         // Holder
@@ -600,7 +600,7 @@ The ``RevokeCredentialEvent`` event is serialized as: first a byte with the valu
 ``IssuerMetadata``
 ^^^^^^^^^^^^^^^^^^
 
-A ``IssuerMetadata`` event MUST be logged when setting the metadata url of the issuer.
+A ``IssuerMetadata`` event MUST be logged when setting the metadata URL of the issuer.
 It consists of a URL for the location of the metadata for the issuer with an optional SHA256 checksum of the content.
 
 The ``IssuerMetadata`` event is serialized as: first a byte with the value of 247, followed by :ref:`CIS-2-MetadataUrl` (``metadata``)::
@@ -635,7 +635,7 @@ The ``CredentialSchemaRefEvent`` event is serialized as: first a byte with the v
 ^^^^^^^^^^^^^^^^^^^^^^
 
 A ``RevocationKeyEvent`` event MUST be logged when registering a new or removing an existing revocation key.
-It consist of the key and the action performed with the key (registration or removal).
+It consists of the key and the action performed with the key (registration or removal).
 
 The ``RevocationKeyEvent`` event is serialized as: first a byte with the value of 244, followed by the key bytes :ref:`CIS-4-PublicKeyEd25519` and 1 byte encoding the action (0 for ``Register``, 1 for ``Remove``)::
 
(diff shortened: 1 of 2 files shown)


source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+It is serialized as :ref:`CIS-4-CredentialInfo` (``credential_info``) followed by a credential schema reference :ref:`CIS-4-SchemaRef` (``schema_ref``), and a credential entry revocation nonce :ref:`CIS-4-Nonce` (``revocation_nonce``)::
+
+  CredentialQueryResponse ::= (credential_info: CredentialInfo) (schema_ref: SchemaRef) (revocation_nonce: Nonce)
+
+
+Requirements
+~~~~~~~~~~~~
+
+- The query MUST fail if the credential ID is unknown.
```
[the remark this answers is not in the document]
Bargsteen:
```suggestion
- The query MUST fail if the credential ID is not present in the registry.
```
Reuse the phrasing from further below. `unknown` can be ambiguous.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+      The smart contract logic SHOULD practice its best efforts to ensure that only the holder can generate and authorize a revocation request with a valid signature.
+
+.. _CIS-4-functions-revokeCredentialOther:
+
+``revokeCredentialOther``
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Revoke a credential by a revocation authority request.
+A revocation authority is any entity that holds a private key corresponding to the public key registered by the issuer.
+A revocation authority is authorized to revoke a credential by verifying the signature with the public key this the given identifier.
```
Bargsteen, line 428: Something is off towards the end of this sentence.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+This entrypoint gives a general way of adding revocation rights to external entities.
+It replaces the authorization checks conducted on the ``sender/invoker`` variable with signature verification.
+In particular, it enables the issuer to provide a service for selected entities to revoke credentials without paying for revocation transactions.
+
+
+Parameter
+~~~~~~~~~
+
+It is serialized as :ref:`CIS-4-SignatureEd25519` (``signature``) and message data ``RevocationDataOther`` consisting of :ref:`CIS-4-CredentialID` (``credential_id``), metadata about the signature :ref:`CIS-4-SigningData` (``signing_data``), a reovocation public key :ref:`CIS-4-PublicKeyEd25519` , and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`::
```
Bargsteen, line 438:
```suggestion
It is serialized as :ref:`CIS-4-SignatureEd25519` (``signature``) and message data ``RevocationDataOther`` consisting of :ref:`CIS-4-CredentialID` (``credential_id``), metadata about the signature :ref:`CIS-4-SigningData` (``signing_data``), a revocation public key :ref:`CIS-4-PublicKeyEd25519` , and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`::
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+Requirements
+~~~~~~~~~~~~
+
+- If revoked successfully, the credential status MUST change to ``Revoked`` (see :ref:`CIS-4-functions-credentialStatus`).
+- The ``RevokeCredentialOtherParam``'s ``signing_data`` MUST include a nonce to protect against replay attacks.
+  The revocation authority's nonce is sequentially increased every time a revocation request is successfully executed.
+  The function MUST only accept a ``RevokeCredentialOtherParam`` if it has the next nonce following the sequential order.
+- The revocation MUST fail if:
+    - The credential ID is not present in the registry.
+    - The revocation key in unknown.
```
Bargsteen, line 453: Can we use the not present phrasing here as well?

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+It is serialized as: First 2 bytes encode the length (``n``) the vector of keys, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
+
+  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ
+
+Requirements
+~~~~~~~~~~~~
+
+- The revocation MUST fail if:
+    - The the transaction ``sender`` is not the issuer.
+    - Some of the keys are unknown.
```
Bargsteen, line 504: Present in.. ;)

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+A custom event SHOULD NOT have a first byte colliding with any of the events defined by this specification.
+
+.. _CIS-4-register-credential-transfer:
+
+``RegisterCredentialEvent``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A ``RegisterCredentialEvent`` event MUST be logged when a new credential is registered with :ref:`CIS-4-functions-registerCredential`.
+
+The ``RegisterCredentialEvent`` event is serialized as: a first a byte with the value of 255, followed by :ref:`CIS-4-CredentialID` (``crednetial_id``), a public key of the holder :ref:`CIS-4-PublicKeyEd25519` (``holder_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
```
Bargsteen, line 538:
```suggestion
The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 255, followed by :ref:`CIS-4-CredentialID` (``crednetial_id``), a public key of the holder :ref:`CIS-4-PublicKeyEd25519` (``holder_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
```

Bargsteen (changes requested):
Nice work!
I found a number of smaller typos etc., but the primary findings are regarding some of the requirements.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+  The holders's nonce is sequentially increased every time a revocation request is successfully executed.
+  The function MUST only accept a ``RevokeCredentialHolderParam`` if it has the next nonce following the sequential order.
+- The revocation MUST fail if:
+    - The credential ID is not present in the registry.
+    - The credential status is not one of ``Active`` or ``NotActivated`` (see :ref:`CIS-4-functions-credentialStatus`).
+    - The credential is not holder-revocable.
+    - The signature was intended for a different contract.
+    - The signature was intended for a different entrypoint.
+    - The signature is expired.
+    - The signature can not be validated.
```
limemloh, line 418:
```suggestion
    - The signature cannot be validated.
```

source/CIS/cis-4.rst
limemloh, line 428:
```suggestion
A revocation authority is authorized to revoke a credential by verifying the signature with the public key of the given identifier.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+.. _CIS-4-functions-revokeCredentialHolder:
+
+``revokeCredentialHolder``
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Revoke a credential by the holders's request.
+
+The holder is authorized to revoke a credential by verifying the signature with the holder's public key.
+It replaces the authorization checks conducted on the ``sender/invoker`` variable with signature verification.
```
limemloh, line 392: What is the `sender/invoker` variable? Are you referring to the sender of the contract update? If so, consider making this more explicit, ideally this standard would make sense to people who are not familiar to our smart contract Rust API

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+Requirements
+~~~~~~~~~~~~
+
+- If revoked successfully, the credential status MUST change to ``Revoked`` (see :ref:`CIS-4-functions-credentialStatus`).
+- The ``RevokeCredentialHolderParam``'s ``signing_data`` MUST include a nonce to protect against replay attacks.
+  The holders's nonce is sequentially increased every time a revocation request is successfully executed.
+  The function MUST only accept a ``RevokeCredentialHolderParam`` if it has the next nonce following the sequential order.
+- The revocation MUST fail if:
+    - The credential ID is not present in the registry.
```
limemloh, line 412: Again consider adding the rejection codes to the standard

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+It is serialized as First 2 bytes encode the length (``n``) the vector of kesy, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
+
+  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ
+
+Requirements
+~~~~~~~~~~~~
+
+- The revocation MUST fail if:
+    - The the transaction ``sender`` is not the issuer.
```
limemloh, line 479:
```suggestion
    - The sender of the transaction is not the issuer.
```

source/CIS/cis-4.rst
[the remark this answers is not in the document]
annenkov: > Should we also state that it must return the info if the id is present?

I don't think we need to explicitly say it in `Requirements`, the `Response` section states what is returned.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+Response
+~~~~~~~~
+
+The function output is the issuer's account address.
+It is serialized as :ref:`CIS-4-AccountAddress`.
+
+.. _CIS-4-functions-issuerMetadata:
+
+``issuerMetadata``
```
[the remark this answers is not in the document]
annenkov: This entrypoint doesn't have a parameter. This is metadata for the whole register, since it's a single issuer per register.
(There is a different kind of metadata: credential metadata. To query that one you need a credential id)

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+^^^^^^^^^^^^^^^^^^
+
+Query the current token metadata URLs for a list of token IDs.
+
+Response
+~~~~~~~~
+
+The function output is the issuer's metadata URL.
+
+It is serialized as :ref:`CIS-2-MetadataUrl`.
```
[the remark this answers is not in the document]
annenkov: No, it's just a single `MetadataUrl` value for the issuer.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
annenkov: I'll change the phrasing "unknown" -> "not present in the registry" in all the relevant cases.

source/CIS/cis-4.rst
annenkov: Done.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+.. TODO: what kind of characters are allowed? ASCII, Unicode?
+
+Requirements
+~~~~~~~~~~~~
+
+- If revoked successfully, the credential status MUST change to ``Revoked`` (see :ref:`CIS-4-functions-credentialStatus`).
+- The revocation MUST fail if:
+    - The the transaction ``sender`` is not the issuer.
+    - The credential ID is not present in the registry.
```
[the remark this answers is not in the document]
annenkov: I've considered this in the beginning. After discussing it with @abizjak, it looks like requiring certain rejection codes restricts implementation. One needs to specify the exact situations when the codes are returned in order for this to be reliably used in the wallet.
After that, I've decided to not specify any rejection codes.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+It is serialized as: First byte encodes the length (``n``) of the credential type, followed by this many bytes for the credential type string::
+
+  CredentialType ::= (n: Byte) (credential_type: Byteⁿ)
+
+.. _CIS-4-Commitment:
+
+``Commitment``
+^^^^^^^^^^^^^^
+
+A vector Pedersen commitment to the credential attributes.
```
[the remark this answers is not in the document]
annenkov: This was one of the possibilities, but in the science team we picked "vector Pedersen commitment" and it is used in this form in the blue paper. Let's stick with it here as well.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+The function output a list of available revocation keys.
+Valid signatures with the corresponding private keys can be used to revoke any credential in the registry.
+
+It is serialized as: First 2 bytes encode the length (``n``) the vector of keys, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
+
+  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ
+
+
+Logged events
```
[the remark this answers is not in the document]
annenkov: Yes, this is a good point.
I think @abizjak was against it, but I don't remember why :)

source/CIS/cis-4.rst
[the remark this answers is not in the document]
Bargsteen: Ok, fair enough. 👍

source/CIS/cis-4.rst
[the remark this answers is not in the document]
Bargsteen: Ah, my mistake.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
Bargsteen: Ok

source/CIS/cis-4.rst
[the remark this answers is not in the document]
abizjak: You mean not to make them clash with tags from other standards?

While it's true you cannot make CIS-2 + CIS-4 supporting contract, it is hard to imagine such a use case.

Having to have all distinct tags from different standards is hopeless.

I don't see this being a problem and I don't want to change things yet again.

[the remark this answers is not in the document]
abizjak: If this was a concern the only real solution would be to include the name of the standard in the event.

So we'd have events

```
struct Event {
   standard: Standard,
   standard_event: StandardDefinedEvent
}
```
where `Standard` would be names, so `CIS-0`, `CIS-1`, ...
and would be emitting these instead of requiring a globally unique u8 tag.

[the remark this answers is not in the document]
abizjak: Or instead of the full `CIS-?` it could just be a u16 or u32 with the standard ID.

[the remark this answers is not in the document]
abizjak: If you think there is a case where this will be a problem we can still change it now, it's just annoying since there are many components.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
abizjak: This is based on the experience of using CIS2 contracts.

The codes are not used because they are not reliable in general. So it's hard to make use of them in a meaningful way, e.g., in a client that interacts with general CIS2 contracts.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
abizjak: There is also CIS3 which has this 250 event for nonce. It is conceivable to sponsor issuance perhaps so we need to then not use 250.

I don't want to spend more time revising serialization so we'll then just use 249.. codes for now.

But any future standard we must revise this, because it is not tenable to cram all possible event types into a u8 globally.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+  Revoker ::= (0: Byte)                         // Issuer
+            | (1: Byte)                         // Holder
+            | (2: Byte) (key: PublicKeyEd25519) // Other
+  RevokeCredentialEvent ::= (254: Byte) (credential_id: CredentialID) (holder_id: PublicKeyEd25519) (revoker: Revoker) (reason: OptionReason)
+
+``IssuerMetadata``
+^^^^^^^^^^^^^^^^^^
+
+A ``IssuerMetadata`` event MUST be logged when updating the credential metadata.
+This also applies to contract initialization.
```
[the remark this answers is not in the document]
abizjak: What is your comment here? That we don't want to necessarily emit the metadata event in initialization?

If we require it then it is simple to know the metadata from events by just looking at events. THere is no special casing needed. Or is your comment about something else?

source/CIS/cis-4.rst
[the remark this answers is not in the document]
abizjak: The issue being that the codes are only defined in the sense of "if code is X then this must have happened", but we do not prescribe an `if and only if` since it prescribes implementation too much. For example our contracts often fail with "serialization failure".

source/CIS/cis-4.rst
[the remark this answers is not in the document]
annenkov: I agree with @abizjak, I think it is useful to log the event in initialization. @limemloh what do you think? Should it be phrased differently?

source/CIS/cis-4.rst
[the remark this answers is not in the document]
annenkov: I like the idea of using standard ids as tags in future CIS  standards.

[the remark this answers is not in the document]
annenkov: I fixed the event tags. Now they start with 249 counting down.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+``CredentialMetadataEvent``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A ``CredentialMetadataEvent`` event MUST be logged when updating the credential metadata.
+It consist of a credential ID and a URL for the location of the metadata for this credential with an optional SHA256 checksum of the content.
+
+The ``TokenMetadataEvent`` event is serialized as: first a byte with the value of 252, followed by the token ID :ref:`CIS-2-TokenID` (``id``), and then a :ref:`CIS-2-MetadataUrl` (``metadata``)::
+
+  TokenMetadataEvent ::= (252: Byte) (id: CredentialID) (metadata: MetadataUrl)
```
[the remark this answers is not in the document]
annenkov: Fixed.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+It is serialized as :ref:`CIS-4-CredentialID` (``credential_id``) followed by :ref:`CIS-4-CredentialInfo` (``credential_info``)::
+
+  RegisterCredentialParameter ::= (credential_id: CredentialID) (credential_info: CredentialInfo)
+
+Requirements
+~~~~~~~~~~~~
+
+- The credential registration request MUST fail if the credential ID is already present in the registry.
+- The credential status after registration MUST NOT be ``Revoked`` (See the possible values for the status in :ref:`CIS-4-CredentialStatus`).
```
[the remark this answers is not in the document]
annenkov: Fixed.

source/CIS/cis-4.rst
annenkov: Fixed.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
limemloh: Right, this was not very clear. The event is good to have, but I don't think we should have a requirement for it to be emitted during contract initialization.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
limemloh: yes, we should have a better solution in the future

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+
+.. _CIS-4-functions-revokeCredentialHolder:
+
+``revokeCredentialHolder``
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Revoke a credential by the holders's request.
+
+The holder is authorized to revoke a credential by verifying the signature with the holder's public key.
+It replaces the authorization based on checking the transaction sender address with signature verification.
```
abizjak, line 400: Why are we not specifying the exact message to be signed, with the domain separation string included?

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+
+
+Requirements
+~~~~~~~~~~~~
+
+- If revoked successfully, the credential status MUST change to ``Revoked`` (see :ref:`CIS-4-functions-credentialStatus`).
+- The ``RevokeCredentialHolderParam``'s ``signing_data`` MUST include a nonce to protect against replay attacks.
+  The holders's nonce is sequentially increased every time a revocation request is successfully executed.
+  The function MUST only accept a ``RevokeCredentialHolderParam`` if it has the next nonce following the sequential order.
+- The revocation MUST fail if:
```
abizjak, line 419: Make it clear that the list is an "or"., i.e., `must fail if any of the following conditions are met`.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+
+``revocationKeys``
+^^^^^^^^^^^^^^^^^^
+
+Query revocation keys.
+
+Response
+~~~~~~~~
+
+The function output a list of available revocation keys.
```
abizjak, line 525:
```suggestion
The function outputs a list of available revocation keys.
```

source/CIS/cis-4.rst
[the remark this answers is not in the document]
limemloh: I agree, that we don't want to add requirements for the error codes, but standardizing an error code and the meaning of it, is still useful. If you want your contract to give a good user experience, you would want to provide the right error information back to the user and having some standardized error codes for this helps.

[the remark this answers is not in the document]
abizjak: What I tried to write above is that I don't think it actually helps.

But if you want to do it I'm not against it, but for me there isn't much value in this rpecisely because you cannot rely on it.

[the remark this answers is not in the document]
limemloh: An application should not rely their logic on this, no. But in the case of a UI it is useful, the wallet could show a nice error message and the contract developer would be motivated to take advantage of this.
I guess the schema for error codes, sort of solves this in the general case.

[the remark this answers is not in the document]
limemloh: This is just a consideration

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+
+
+.. _CIS-4-CredentialStatus:
+
+``CredentialStatus``
+^^^^^^^^^^^^^^^^^^^^
+
+The status of a verifiable credential.
+
+It is serialized as 1 byte where ``0`` correponds to the status ``Active``, ``1`` corresponds to  ``Revoked``, ``2`` corresponds to  ``Expired``, ``3`` corresponds to ``NotActivated``::
```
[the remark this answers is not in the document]
annenkov: I explained the meaning of statuses in the requirements for the `credentialStatus` entrypoint. I'll  add reference here.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+
+  RegisterCredentialParameter ::= (credential_id: CredentialHolderId) (credential_info: CredentialInfo)
+
+Requirements
+~~~~~~~~~~~~
+
+- The credential registration request MUST fail if the credential ID is already present in the registry.
+- After successful registration:
+    - Querying the credential by its ID with :ref:`CIS-4-functions-credentialEntry` MUST succeed.
+    - Querying the credential status by ID with :ref:`CIS-4-functions-credentialStatus` MUST succeed and MUST NOT return ``Revoked`` (See the possible values for the status in :ref:`CIS-4-CredentialStatus`).
```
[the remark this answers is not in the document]
annenkov: My reasoning was that revocation must always be explicit generating a revocation event. In this case, it's clear that if there is no revocation event in the log, the credential is not revoked. However, it was before we introduced the possibility of restoring revoked credentials. With the restoration functionality, a revocation event does not mean that the credential stays revoked. Indeed, it seems like it's not necessary to specify this.

source/CIS/cis-4.rst
annenkov: Yes, good point. I'll add the message format.

source/CIS/cis-4.rst
annenkov: Done.

source/CIS/cis-4.rst
annenkov: Fixed.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+The parameter is the credential ID :ref:`CIS-4-CredentialHolderId` and optional string indicating the revocation reason.
+
+It is serialized as :ref:`CIS-4-CredentialHolderId` followed by 1 byte to indicate whether a reason is included.
+If its value is 0, then no reason string is present, if the value is 1 then the bytes corresponding to the reason string follow::
+
+  OptionalReason ::= (0: Byte)
+                   | (1: Byte) (n: Byte) (reason_string: Byteⁿ)
+  RevokeCredentialIssuerParam ::= (credential_id: CredentialHolderId) (reason: OptionReason)
+
+.. TODO: what kind of characters are allowed? ASCII, Unicode?
```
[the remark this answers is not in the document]
annenkov: Done.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,733 @@
...
+
+Serialized in the same way as :ref:`CIS-2 MetadataUrl<CIS-2-MetadataUrl>`.
+
+
+.. _CIS-4-CredentialType:
+
+``CredentialType``
+^^^^^^^^^^^^^^^^^^
+
+Is a short ASCII string (up to 256 characters) describing the credential type that is used to identify which schema the credential is based on.
```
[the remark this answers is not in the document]
annenkov: Changed to UTF-8.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,705 @@
...
+
+The parameter is the credential ID :ref:`CIS-4-CredentialID` and optional string indicating the revocation reason.
+
+It is serialized as :ref:`CIS-4-CredentialID` followed by 1 byte to indicate whether a reason is included, if its value is 0, then no reason string present, if the value is 1 then the bytes corresponding to the reason string follow::
+
+  OptionReason ::= (0: Byte)
+                 | (1: Byte) (n: Byte) (reason_string: Byteⁿ)
+  RevokeCredentialIssuerParam ::= (credential_id: CredentialID) (reason: OptionReason)
+
+.. TODO: what kind of characters are allowed? ASCII, Unicode?
```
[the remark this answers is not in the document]
annenkov: Fixed TODO, use UTF-8.

source/CIS/cis-4.rst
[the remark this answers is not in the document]
annenkov: It's hard for me to see which codes to standardize and which not. Issuance is usually issuer-specific, not much to standardize there; verifiers and holders (wallets) mostly fetch data, so, the only relevant error is `CredentialNotFound`.
Holders can revoke in the wallet, but it can fail for multiple reasons, and it's hard to fix the granularity in advance.

This essentially leaves us with  `CredentialNotFound`, which I can include in the standard. It would require a slight change in the credential registry example:  the order of enum constructors [here](https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/credential-registry/src/lib.rs#L172) to make the CIS-4 error have the first tag.
And, of course, we should provide a library in the future similar to CIS-2 that will help avoid error code clashes.

@abizjak @limemloh How does that sound?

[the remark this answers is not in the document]
annenkov: > the order of enum constructors [here](https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/credential-registry/src/lib.rs#L172) to make the CIS-4 error have the first tag.

And I'll probably need to do something similar to the CIS-2 library https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/concordium-cis2/src/lib.rs#L923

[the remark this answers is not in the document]
annenkov: I experimented with adding a specific rejection code to the implementation and it looks like it brings more issues than solves. One should make sure that for the `CredentialNotFound` case a specific rejection code is produced and all custom errors don't clash with it. Since there is no clear benefit from doing this, I'll drop this idea and settle on not including any rejection code requirements.

source/CIS/cis-4.rst
```diff
@@ -362,22 +381,21 @@ The issuer is authorized to revoke a credential if the transaction sender's addr
...
-  OptionReason ::= (0: Byte)
-                 | (1: Byte) (n: Byte) (reason_string: Byteⁿ)
-  RevokeCredentialIssuerParam ::= (credential_id: CredentialID) (reason: OptionReason)
+It is serialized as :ref:`CIS-4-CredentialHolderId` followed by 1 byte to indicate whether a reason is included.
+If its value is 0, then no reason string is present, if the value is 1 then the bytes corresponding to the reason string follow::
 
-.. TODO: what kind of characters are allowed? ASCII, Unicode?
+  OptionalReason ::= (0: Byte)
+                   | (1: Byte) (n: Byte) (reason_string: Byteⁿ)
+  RevokeCredentialIssuerParam ::= (credential_id: CredentialHolderId) (reason: OptionReason)
```
Bargsteen, line 391:
```suggestion
  RevokeCredentialIssuerParam ::= (credential_id: CredentialHolderId) (reason: OptionalReason)
```

source/CIS/cis-4.rst
```diff
@@ -435,22 +456,25 @@ In particular, it enables the issuer to provide a service for selected entities
...
 
 - If revoked successfully, the credential status MUST change to ``Revoked`` (see :ref:`CIS-4-functions-credentialStatus`).
+- The message to be signed is produced in the following way:
+    - Start with the bytes of the domain separation string ``WEB3ID:REVOKE``.
+    - Append ``RevocationDataOther`` bytes from the input parameter.
 - The ``RevokeCredentialOtherParam``'s ``signing_data`` MUST include a nonce to protect against replay attacks.
   The revocation authority's nonce is sequentially increased every time a revocation request is successfully executed.
   The function MUST only accept a ``RevokeCredentialOtherParam`` if it has the next nonce following the sequential order.
-- The revocation MUST fail if:
+- The revocation MUST fail if any of the following conditions are met any of the following conditions are met:
```
Bargsteen, line 475:
```suggestion
- The revocation MUST fail if any of the following conditions are met:
```

source/CIS/cis-4.rst
```diff
@@ -533,45 +559,52 @@ A custom event SHOULD NOT have a first byte colliding with any of the events def
...
+The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 254, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`; ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
 
   Revoker ::= (0: Byte)                         // Issuer
             | (1: Byte)                         // Holder
             | (2: Byte) (key: PublicKeyEd25519) // Other
-  RevokeCredentialEvent ::= (254: Byte) (credential_id: CredentialID) (holder_id: PublicKeyEd25519) (revoker: Revoker) (reason: OptionReason)
+  RevokeCredentialEvent ::= (248: Byte) (credential_id: CredentialHolderId) (revoker: Revoker) (reason: OptionReason)
```
Bargsteen, line 581:
```suggestion
  RevokeCredentialEvent ::= (248: Byte) (credential_id: CredentialHolderId) (revoker: Revoker) (reason: OptionalReason)
```

Bargsteen (approved):
LGTM

limemloh (approved):

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,750 @@
...
+
+ {
+    "title": "Concordium Employment",
+    "logo" : {
+      "url":  "https://concordium.com/wp-content/uploads/2022/07/Concordium-1.png",
+      "hash": "1c74f7eb1b3343a5834e60e9a8fce277f2c7553112accd42e63fae7a09e0caf8"
+      }
+    "background_color": "#000000",
+    "image": {
+      "url": "https://concordium.com/employment/vc-background.png",
```
orhoj, line 735: There is a hash for the `logo` but not for the `image`. Should it be the same?

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,750 @@
...
+.. list-table:: Credential metadata JSON Object
+  :header-rows: 1
+
+  * - Property
+    - JSON value type [JSON-Schema]
+    - Description
+  * - ``title``
+    - string
+    - The name to display for the credential.
+  * - ``logo``
```
orhoj, line 707: Do we want to specify formats (sizes) in the standard document? Or do we want to do that elsewhere?

source/CIS/cis-4.rst
annenkov: `hash` is optional, so I wanted to give an example with and without `hash`. I can add a comment explaining this.

source/CIS/cis-4.rst
annenkov: Good question. CIS-2 does not specify the sizes. I think it makes sense to leave them out. CIS standards are about the API, additional details might be specified elsewhere.

@orhoj is there a document that specifies the image sizes for CIS-2 complaint NFTs?

annenkov: Let's add a recommendation about the image sizes to the standard. This recommendation might be revised later depending on the final version of the wallet's UI.

annenkov: Done. I added initial values and a TODO to check before we finalize the standard.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,787 @@
...
+
+.. _CIS-4-register-credential-transfer:
+
+``RegisterCredentialEvent``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A ``RegisterCredentialEvent`` event MUST be logged when a new credential is registered.
+The event records the credential identifier, the credential type, and the corresponding schema reference.
+
+The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 255, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
```
limemloh, line 575:
```suggestion
The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 249, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,787 @@
...
+  CredentialEventData ::= (credential_id: CredentialHolderId) (schema_ref: SchemaRef) (credential_type: CredentialType)
+  RegisterCredentialEvent ::= (249: Byte) (data: CredentialEventData)
+
+``RevokeCredentialEvent``
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A ``RevokeCredentialEvent`` event MUST be logged when a credential is revoked.
+The event records the credential identifier, who requested the revocation (the holder, the issuer or a revocation authority), and an optional string with a short comment on the revocation reason.
+
+The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 254, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`; ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
```
limemloh, line 586:
```suggestion
The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 248, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`; ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,787 @@
...
+
+.. _CIS-4-events-IssuerMetadata:
+
+``IssuerMetadata``
+^^^^^^^^^^^^^^^^^^
+
+A ``IssuerMetadata`` event MUST be logged when setting the metadata url of the issuer.
+It consists of a URL for the location of the metadata for the issuer with an optional SHA256 checksum of the content.
+
+The ``IssuerMetadata`` event is serialized as: first a byte with the value of 253, followed by :ref:`CIS-2-MetadataUrl` (``metadata``)::
```
limemloh, line 602:
```suggestion
The ``IssuerMetadata`` event is serialized as: first a byte with the value of 247, followed by :ref:`CIS-2-MetadataUrl` (``metadata``)::
```

annenkov:
Thanks @limemloh, nice catch about the event tags! :)

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,788 @@
...
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+
+Parameter
+~~~~~~~~~
+
+It is serialized as First 2 bytes encode the length (``n``) of the vector of keys, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
+
+  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ
```
soerenbf, line 508: In the `credential-registry` contract example, this also includes auxiliary data (I see you included it in the note, so maybe you just forgot to add it here as well?)
```suggestion
  AuxData ::= (n: Byte²) (data: Byteⁿ)
  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ (auxiliary_data: AuxData)
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,788 @@
...
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+
+Parameter
+~~~~~~~~~
+
+It is serialized as: First 2 bytes encode the length (``n``) of the vector of keys, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
+
+  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ
```
soerenbf, line 536: And here:
```suggestion
  AuxData ::= (n: Byte²) (data: Byteⁿ)
  RegisterPublicKeyParameters ::= (n: Byte²) (key: PublicKeyEd25519)ⁿ (auxiliary_data: AuxData)
```

soerenbf (commented):
Noticed more discrepancies when comparing to the example implementation 😄

source/CIS/cis-4.rst
annenkov: Yep, nice catch! Thanks!

source/CIS/cis-4.rst
annenkov: Indeed, thanks!

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,796 @@
...
+  SigningData ::= (contract_address: ContractAddress) (entrypoint: EntrypointName) (nonce: Nonce) (timestamp: Timestamp)
+
+.. _CIS-4-SchemaRef:
+
+``SchemaRef``
+^^^^^^^^^^^^^
+
+A URL of the credential schema.
+
+Serialized in the same way as :ref:`CIS-2 MetadataUrl<CIS-2-MetadataUrl>`.
```
soerenbf, line 185: Do we maybe want to link to the CIS4 metadataURL serialization instead? It seems odd to me that we duplicate the serialized format and then link to the CIS2 version here. Linking to the CIS4 reference also helps the user stay in the document instead of jumping back and forth.

```suggestion
Serialized in the same way as :ref:`CIS-4-MetadataUrl`.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,796 @@
...
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+
+Parameter
+~~~~~~~~~
+
+The parameter is the credential ID :ref:`CIS-4-CredentialHolderId` and an optional string in the UTF-8 encoding that indicates the revocation reason.
+
+It is serialized as :ref:`CIS-4-CredentialHolderId` followed by 1 byte to indicate whether a reason is included.
+If its value is 0, then no reason string is present, if the value is 1 then the bytes corresponding to the reason string follow::
```
soerenbf, line 393: The auxiliary data part of the parameter is left out here. Is that on purpose?

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,796 @@
...
+.. note::
+  This standard does not specify how the issuer is authenticated.
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+
+Parameter
+~~~~~~~~~
+
+It is serialized as: First 2 bytes encode the length (``n``) of the vector of keys, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
```
soerenbf, line 541: Also here the aux data is left out (ignore if it's on purpose)

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,796 @@
...
+  This standard does not specify how the issuer is authenticated.
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
+
+Parameter
+~~~~~~~~~
+
+It is serialized as First 2 bytes encode the length (``n``) of the vector of keys, followed by this many :ref:`CIS-4-PublicKeyEd25519` keys::
+
```
soerenbf, line 513: And here

soerenbf (commented):
Looks good to me overall, and corresponds with the CIS4 implementation done in the JS SDK, which has been tested with the credential-registry example.

source/CIS/cis-4.rst
annenkov: Yes, you're right, it should be a local reference to `MetadataUrl`. I see, however, that there is a repetition of common data types in several standards. Maybe (later) we can organize the standards as a library with common definitions being some sort of `std`

source/CIS/cis-4.rst
annenkov: Good catch, `AuxData` should be in the explanation.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+.. _CIS-4-MetadataUrl:
+
+``MetadataUrl``
+^^^^^^^^^^^^^^^
+
+A URL and optional checksum for metadata stored outside of this contract.
+
+It is serialized as: 2 bytes for the length of the metadata url (``n``) and then this many bytes for the url to the metadata (``url``) followed by an optional checksum.
+The checksum is serialized by 1 byte to indicate whether a hash of the metadata is included.
+If its value is 0, then there is no hash, if the value is 1 then 32 bytes for a SHA256 hash (``hash``) follows::
```
dg-concordium, line 80:
```suggestion
If its value is 0, then there is no hash; if the value is 1, then 32 bytes for a SHA256 hash (``hash``) follows::
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+
+
+.. _CIS-4-MetadataUrl:
+
+``MetadataUrl``
+^^^^^^^^^^^^^^^
+
+A URL and optional checksum for metadata stored outside of this contract.
+
+It is serialized as: 2 bytes for the length of the metadata url (``n``) and then this many bytes for the url to the metadata (``url``) followed by an optional checksum.
```
dg-concordium, line 78:
```suggestion
It is serialized as: 2 bytes for the length of the metadata URL (``n``) and then this many bytes for the URL to the metadata (``url``) followed by an optional checksum.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+``registerCredential``
+^^^^^^^^^^^^^^^^^^^^^^
+
+Register public data for a new credential.
+
+.. note::
+  This standard does not specify how the issuer is authenticated.
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
```
dg-concordium, line 356:
```suggestion
  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+
+Parameter
+~~~~~~~~~
+
+The parameter is credential information that is used to create an entry in the registry.
+
+It is serialized as :ref:`CIS-4-CredentialInfo` (``credential_info``) followed by auxiliary data, which is serialized as 2 bytes to encode the length (``n``) of the vector of keys, followed by this many bytes of data::
```
dg-concordium, line 363:
```suggestion
It is serialized as :ref:`CIS-4-CredentialInfo` (``credential_info``), followed by auxiliary data, which is serialized as 2 bytes to encode the length (``n``) of the vector of keys, followed by this many bytes of data::
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+``revokeCredentialIssuer``
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Revoke a credential by the issuer's request.
+
+.. note::
+  This standard does not specify how the issuer is authenticated.
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
```
dg-concordium, line 385:
```suggestion
  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+
+Parameter
+~~~~~~~~~
+
+The parameter is the credential ID :ref:`CIS-4-CredentialHolderId` and an optional string in the UTF-8 encoding that indicates the revocation reason.
+
+It is serialized as :ref:`CIS-4-CredentialHolderId` followed by 1 byte to indicate whether a reason is included.
+If its value is 0, then no reason string is present, if the value is 1 then the bytes corresponding to the reason string follow.
```
dg-concordium, line 393:
```suggestion
If its value is 0, then no reason string is present; if the value is 1, then the bytes corresponding to the reason string follow.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+  The function MUST only accept a ``RevokeCredentialOtherParam`` if it has the next nonce following the sequential order.
+- The revocation MUST fail if any of the following conditions are met:
+    - The credential ID is not present in the registry.
+    - The revocation key in not present in the registry.
+    - The credential status is not one of ``Active`` or ``NotActivated`` (see :ref:`CIS-4-functions-credentialStatus`).
+    - The signature was intended for a different contract.
+    - The signature was intended for a different entrypoint.
+    - The signature is expired.
+    - The signature can not be validated.
+      The smart contract logic SHOULD practice its best efforts to ensure that only the revocation authority can generate and authorize a revocation request with a valid signature.
```
dg-concordium, line 487: Same as above.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+``removeRevocationKeys``
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Make a list of public keys unavailable to revocation authorities.
+
+.. note::
+  This standard does not specify how the issuer is authenticated.
+  Implementations can use various mechanisms.
+  For example, the transaction sender's address is checked against the issuer's account address stored in the contract.
+  Another option is to use ``auxiliary_data`` to implement signature-based authentication mechanism.
```
dg-concordium, line 530:
```suggestion
  Another option is to use ``auxiliary_data`` to implement a signature-based authentication mechanism.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+The ``RegisterCredentialEvent`` event is serialized as: first a byte with the value of 249, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a reference to the credential schema :ref:`CIS-4-SchemaRef` (``schema_ref``), a credential type :ref:`CIS-4-CredentialType` (``credential_type``) ::
+
+  CredentialEventData ::= (credential_id: CredentialHolderId) (schema_ref: SchemaRef) (credential_type: CredentialType)
+  RegisterCredentialEvent ::= (249: Byte) (data: CredentialEventData)
+
+``RevokeCredentialEvent``
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A ``RevokeCredentialEvent`` event MUST be logged when a credential is revoked.
+The event records the credential identifier, who requested the revocation (the holder, the issuer or a revocation authority), and an optional string with a short comment on the revocation reason.
```
dg-concordium, line 588:
```suggestion
The event records the credential identifier who requested the revocation (the holder, the issuer, or a revocation authority), and an optional string with a short comment on the revocation reason.
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+
+The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 248, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`; ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
```
dg-concordium, line 590:
```suggestion
The ``RevokeCredentialEvent`` event is serialized as: first a byte with the value of 248, followed by :ref:`CIS-4-CredentialHolderId` (``crednetial_id``), a ``revoker``, and an optional revocation reason (``reason``), serialized similarly to :ref:`CIS-4-functions-revokeCredentialIssuer`. ``revoker`` is serialized as 1 byte to indicate who sent the revocation request ( 0 - issuer, 1 - holder, 2 -revocation authority); if the first byte is 2, then it is followed by a public key :ref:`CIS-4-PublicKeyEd25519` of the revoker::
```

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+
+The ``CredentialSchemaRefEvent`` event is serialized as: first a byte with the value of 245, followed by :ref:`CIS-4-CredentialType` (``type``), and then a :ref:`CIS-4-SchemaRef` (``schema_ref``)::
+
+  CredentialSchemaRefEvent ::= (245: Byte) (type: CredentialType) (schema_ref: SchemaRef)
+
+``RevocationKeyEvent``
+^^^^^^^^^^^^^^^^^^^^^^
+
+A ``RevocationKeyEvent`` event MUST be logged when registering a new or removing an existing revocation key.
+It consist of the key and the action performed with the key (registration or removal).
```
dg-concordium, line 638:
```suggestion
It consists of the key and the action performed with the key (registration or removal).
```

dg-concordium (approved):

source/CIS/cis-4.rst
td202: Do you want to mention the endianness of `n`?

annenkov: Yes, this is a good point! I wonder why it's not mentioned in CIS-2, where I copied the definition from. And what is the endianness we actually use when serializing/deserializing the size?

annenkov: Clarified with @abizjak, it's little endian. I think I can write a general remark about how the size is serialized, because it's used in several places.

source/CIS/cis-4.rst
```diff
@@ -0,0 +1,791 @@
...
+  The function MUST only accept a ``RevokeCredentialHolderParam`` if it has the next nonce following the sequential order.
+- The revocation MUST fail if any of the following conditions are met:
+    - The credential ID is not present in the registry.
+    - The credential status is not one of ``Active`` or ``NotActivated`` (see :ref:`CIS-4-functions-credentialStatus`).
+    - The credential is not holder-revocable.
+    - The signature was intended for a different contract.
+    - The signature was intended for a different entrypoint.
+    - The signature is expired.
+    - The signature cannot be validated.
+      The smart contract logic SHOULD practice its best efforts to ensure that only the holder can generate and authorize a revocation request with a valid signature.
```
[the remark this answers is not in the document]
annenkov: Done, and the one below, too.

PR #68, opened 2024-12-03
Update P7 timings and details.

td202:
## Purpose

Update details relating to the protocol 7 update.

## Changes

- Concrete timings
- Transaction and block hashes related to update.


chportma (approved):
All good.

PR #74, opened 2025-07-02
Draft PLT standard document.

td202:
## Purpose

Define a standard for protocol-level tokens.

## Changes

Introduce CIS-7 standard.

## Checklist

- [ ] My code follows the style of this project.
- [ ] The code compiles without warnings.
- [ ] I have performed a self-review of the changes.
- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.

### DIFF
--- a/source/CIS/cis-7.rst
+++ b/source/CIS/cis-7.rst
@@ -1,0 +1,576 @@
+.. _CIS-7:
+
+===================================
+CIS-7: Protocol-level Tokens (PLTs)
+===================================
+
+.. list-table::
+   :stub-columns: 1
+
+   * - Created
+     - Jun 25, 2025
+   * - Final
+     - TBD
+   * - Supported versions
+     - | Protocol version 9 or newer
+   * - Standard identifier
+     - ``CIS-7``
+
+
+Abstract
+========
+
+A standard interface for protocol-level tokens (PLTs).
+The interface defines the data formats used when interacting with PLTs.
+These data are encoded in Concise Binary Object Representation (CBOR) as defined in :rfc:`8949`.
+In particular, the interface defines the following:
+
+- Initialization parameters
+- Transactions
+- Events
+- Reject reasons
+- Token state
+- Account state
+
+Introduction
+============
+
+Protocol-level tokens (PLTs) provide chain-native support for tokens (other than CCD) without depending on smart contracts for the implementation.
+Each PLT is governed by a Token Module, which manages the token's lifecycle, including creation, transfer, and burning of tokens.
+There may be a single Token Module that manages all PLTs, or there may be multiple Token Modules, each managing one or more PLTs.
+
+The Token Module defines a number of interfaces for interaction with PLTs.
+At each interface, data is exchanged in in Concise Binary Object Representation (CBOR) format.
+The CBOR format is defined in :rfc:`8949`.
+This document defines schemas for the data exchanged at each interface, including initialization parameters, transactions, events, reject reasons, token state, and account state.
+
+Terminology and Conventions
+---------------------------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in :rfc:`2119`.
+
+The schemas for CBOR data are presented using Concise Data Definition Language (CDDL) as defined in :rfc:`8610`.
+
+Example data values are presented in CBOR diagnostic notation as described in :rfc:`8949`.
+
+Specification
+=============
+
+Except as otherwise specified, the Token Module MUST accept data in preferred serialization as defined in :rfc:`8949`.
+The Token Module SHOULD accept well-formed CBOR data that is not in preferred serialization, but conforms to the required schema.
+The Token Module MUST produce data in preferred serialization, and SHOULD produce data in Deterministically Encoded CBOR.
+
+
+Common Types
+------------
+
+.. _CIS-7-token-amount:
+
+``token-amount``
+^^^^^^^^^^^^^^^^
+::
+
+  token-amount = decfrac
+
+A token amount is represented as a CBOR decimal fraction.
+A decimal fraction is a tagged pair of a base-10 exponent and significand.
+From the CDDL prelude (:rfc:`8610`)::
+
+  decfrac = #6.4([e10: int, m: integer])
+
+For ``token-amount``, as of protocol version 9, the exponent (``e10``) must be between 0 and -255 (inclusive).
+The significand (``m``) must be between 0 and 18446744073709551615 (inclusive).
+
+Each PLT defines the number of decimal places in its representation.
+A ``token-amount`` for a given PLT MUST be expressed with the exponent ``e10`` being the negation of the number of decimals.
+Thus, the following ``token-amount``s are not equivalent::
+
+  4([2, 100])      -- 1.00
+  4([6, 1000000])  -- 1.000000
+
+``memo``
+^^^^^^^^
+::
+
+    memo = raw-memo / cbor-memo
+
+    raw-memo = bytes .size (0..256)
+
+    cbor-memo = #6.24(raw-memo)
+
+A ``memo`` is a byte string of up to 256 bytes.
+The ``memo`` can be represented either directly as a byte string (``raw-memo``), or as a byte string tagged as CBOR-encoded data (``cbor-memo``).
+The tag 24 is defined in :rfc:`8949#section-3.4.5.1` to denote that the enclosed byte string represents CBOR-encoded data.
+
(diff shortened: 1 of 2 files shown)


source/CIS/cis-7.rst
```diff
@@ -0,0 +1,576 @@
...
+
+::
+
+    token-module-state = {
+        ; The name of the token
+        "name": text,
+        ; A URL pointing to the token metadata
+        "metadata": metadata-url,
+        ; The governance account of the token
+        "governanceAccount": token-holder
```
td202, line 525: Should more of these be optional? Especially `governanceAccount` may be less relevant when we introduce role-based access control.

allanbrondum: If you want a trustless bridge of some token, I think it is common to "throw away" the ability to mint and burn via some trusted account. On some chains, you just have to say, yes "I threw away my keys". On other chains, you can verify on chain, that only the bridge can mint tokens. The functionality to lock yourself out from doing ad hoc minting does not have to be at PLT level, it could also be if we can remove all keys for an account (I don't know if this is possible on Concordium)

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,576 @@
...
+
+Introduction
+============
+
+Protocol-level tokens (PLTs) provide chain-native support for tokens (other than CCD) without depending on smart contracts for the implementation.
+Each PLT is governed by a Token Module, which manages the token's lifecycle, including creation, transfer, and burning of tokens.
+There may be a single Token Module that manages all PLTs, or there may be multiple Token Modules, each managing one or more PLTs.
+
+The Token Module defines a number of interfaces for interaction with PLTs.
+At each interface, data is exchanged in in Concise Binary Object Representation (CBOR) format.
```
drsk0, line 43:
```suggestion
At each interface, data is exchanged in Concise Binary Object Representation (CBOR) format.
```

source/CIS/cis-7.rst
andynelson: Yep, that makes sense...

So I think your proposal would be that we make `governanceAccount` optional, leaving only `name` and `metadata`?

Can we see any future where we would not require having a metadata url?
If so then `name` could be the only required field with all the others being optional.
(Since there's no conceivable future where a PLT didn't need a name, I think that's the one that we'll always need to have)

I for some reason thought we also had the module reference here also, or is that only returned when listing all PLTs?

DOBEN (commented):
Our `CIS` standards currently include protocol standards as well as smart contract standards. More of a long-term thought, if we want to separate these two types into protocol interoperability standards `CIS` and smart contract interoperability standards `CRC (equivalent to ERC on Ethereum)`

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+^^^^^^^^^^^^^^^^
+::
+
+  token-amount = decfrac
+
+A token amount is represented as a CBOR decimal fraction.
+A decimal fraction is a tagged pair of a base-10 exponent and significand.
+From the CDDL prelude (:rfc:`8610`)::
+
+  decfrac = #6.4([e10: int, m: integer])
```
allanbrondum, line 79: Can we exclude bignums from the type? And our schema in general. It could simplify decoding without really putting any requirements on encoders (since not using bignums will be simpler for an encoder). Alternatively we need to make sure our CBOR decoders can handle bignums (Rust does not currently).

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+        * simple-key => details-value
+    }
+
+    simple-key = short-text / uint
+
+    value-0 =
+        tagged-account-address  ; An account address
+        / int                   ; An integer
+        / bigint                ; A big integer
+        / decfrac               ; A decimal fraction
```
allanbrondum, line 328: Same here, can we exclude bignums?

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+    }
+
+Token Modules that use a different format for initialization parameters SHOULD represent the parameters in a key-value map.
+Where keys that are the same as those above are used in initialization parameters, their semantics SHOULD be the same or substantially similar.
+
+
+Transactions
+------------
+
+A Token Update transaction identifies a PLT by its Token ID and carries a CBOR-encoded payload that consists of a list of token operations (``token-update-transaction``).
```
allanbrondum, line 205: Can we describe the context in which the CBOR operations are used. Namely in the token update transaction

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+
+    token-module-state = {
+        ; The name of the token
+        "name": text,
+        ; A URL pointing to the token metadata
+        "metadata": metadata-url,
+        ; The governance account of the token
+        ? "governanceAccount": tagged-account-address
+        ; Whether the token supports an allow list.
+        ? "allowList": bool,
```
allanbrondum, line 532: We should add paused?

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+        ? "allowList": bool .default false,
+        ; Whether the token supports a deny list
+        ? "denyList": bool .default false,
+        ; The initial supply of the token. If not present, no tokens are minted initially.
+        ? "initialSupply": token-amount,
+        ; Whether the token is mintable
+        ? "mintable": bool .default false,
+        ; Whether the token is burnable
+        ? "burnable": bool .default false
+    }
```
allanbrondum, line 196: Should initialization parameters allow `* text => any`?

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+        ; The maximum representable token amount.
+        "maxRepresentableAmount": token-amount,
+    }
+
+Token State
+-----------
+
+The Token Module state is a representation of the global state of a PLT, which is maintained by the Token Module.
+It is returned as part of a `GetTokenInfo` query.
+The Token Module state does not include state that is managed by the Token Kernel, such as the token identifier and global supply.
```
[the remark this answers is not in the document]
td202: I've added it to the introduction.

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,576 @@
...
+        "url": text,
+        ; An optional sha256 checksum value tied to the content of the URL
+        ? "checksumSha256": sha256-hash
+        ; Additional fields may be included for future extensibility, e.g. another hash algorithm.
+        * text => any
+    }
+
+    sha256-hash = bytes .size(32)
+
+A ``metadata-url`` encodes a URL that identifies metadata, together with an optional sha256 checksum of the contents of the metadata.
```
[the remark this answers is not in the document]
td202: Added.

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,576 @@
...
+
+Specification
+=============
+
+Except as otherwise specified, the Token Module MUST accept data in preferred serialization as defined in :rfc:`8949`.
+The Token Module SHOULD accept well-formed CBOR data that is not in preferred serialization, but conforms to the required schema.
+The Token Module MUST produce data in preferred serialization, and SHOULD produce data in Deterministically Encoded CBOR.
+
+
+Common Types
```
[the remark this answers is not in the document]
td202: Added

tschudid (changes requested):
I have left some comments about the token-format and minor editorial suggestions.

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,944 @@
...
+        3: bytes .size 32
+    }
+
+    ; A subtype of the tagged-coininfo type from BCR-2020-007
+    tagged-ccd-coininfo = #6.40305(ccd-coininfo)
+
+    ccd-coininfo = { 
+        ; The type (1) field is the SLIP44 code for Concordium
+        1: 919 
+        ; The network (2) field is not supported.
```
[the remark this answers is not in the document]
tschudid: I am bit worried that this leads to a situation where a user confuses testnet PLTs for mainnet ones.
On the other hand, as long as wallets clearly show which network they are connected to this should not become an issue.

source/CIS/cis-7.rst
td202: I don't think an account is a particularly useful indicator of the owner/issuer in itself. I have made everything in the state optional so as not to bind future implementations.

source/CIS/cis-7.rst
```diff
@@ -0,0 +1,581 @@
...
+- `undefined`: The undefined value.
+- `list-0`: A list of values the above simple values.
+- `map-0`: A map of simple keys to simple values.
+
+
+Events
+------
+
+The Token Module may emit Token Module Events as a consequence of transaction execution.
+These events are in addition to the ``TokenTransfer``, ``TokenMint``, ``TokenBurn`` and ``TokenCreated`` events, and the semanitcs is dependent on the Token Module implementation.
```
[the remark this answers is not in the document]
td202: There is now a reference to the GRPC API spec.

PR #84, opened 2026-05-25
Add CIS-8 (External Key Registry), CIS-8004 (Agent Registry), and CIS-10 (Asserted External Identifier Registry) specs

MarirosansP:
## Summary

Adds two new CIS standards to `source/CIS/` and registers them in the toctree:

- **CIS-8 — External Key Registry.** Smart-contract interface for recording cryptographically-proven bindings between Concordium accounts and external blockchain public keys. Four proof schemes specified: `ethereum-personal-sign`, `solana-ed25519`, `cosmos-secp256k1`, `fetch-ai-ed25519`. Domain-separated canonical message pinned to `(contract_address, genesis_hash)`. Single-account admin with `host.upgrade` support.

- **CIS-8004 — Agent Registry.** Smart-contract interface for an Agent Registry that records the identity of autonomous software agents as transferable CIS-2 NFTs. Each agent record carries an `agent_uri`, a `metadata_hash`, an optional `external_reference` (pointing at a CIS-8 or CIS-10 binding via a tagged-union `ExternalRefKind`), an independently-lifecycled `agent_wallet` (dual-authorised via a 26-byte `CIS-8004/v1/setAgentWallet` signed message), an open key/value metadata map, and `agent_wallet` + `external_reference` resets on every CIS-2 transfer. Implements the Identity Registry component of [ERC-8004 v2 (Trustless Agents)](https://eips.ethereum.org/EIPS/eip-8004); Reputation and Validation Registry components are out of scope for this version.

CIS-8004 declares CIS-8 as a `Related` standard and cross-references it in its Cross-Contract Verification section. Both are submitted as a single PR because they share the `source/index.rst` toctree update.

## Reference deployments (Concordium-owned, live on chain)

| Network | CIS-8 | CIS-8004 |
|---|---|---|
| DevNet P-11 | `<23, 0>` | `<24, 0>` |
| TestNet | `<12794, 0>` | `<12795, 0>` |

Both DevNet contracts redeployed 2026-05-25; module bytes identical between DevNet and TestNet. Mainnet deployment pending.

## Reviewer notes

- **ERC-8004 alignment.** The CIS-8004 spec lists deliberate departures from ERC-8004 v2 in `Departures from ERC-8004 v2` (informative section): native `revoke`, transfer-resets `agent_wallet` and `external_reference`, identifier is the canonical CIS-2 token address (Base58Check) rather than URI/DID.
- **Cosmos ADR-036 envelope.** Per `cosmos-secp256k1` scheme: this standard does not mandate the exact envelope bytes; deployers MUST pin a choice and document it per instance. Happy to tighten if the team wants a single canonical envelope across deployments.
- **CIS-10 dependency.** CIS-8004 references CIS-10 (verifier-asserted external identifier registry) as the binding target for `ExternalRefKind::Cis10(_)`. CIS-10 is out of scope for this PR; spec text acknowledges that deployments without a CIS-10 companion contract must reject every `Cis10` external reference. The new `setExternalRegistries` admin entrypoint (added in this redeploy) lets the admin re-point the linked CIS-10 address once a CIS-10 instance is deployed, without needing to redeploy CIS-8004.
- **Token ID wire format (CIS-8004).** Spec normatively requires `TokenIdU64` (`[u8 length=8][8 LE bytes]`) on every entrypoint — both CIS-8004-native and CIS-2 inherited. An earlier pre-2026-05-25 deployment of the reference contract incorrectly typed native entrypoints as bare `u64` LE; an informational note in `Token ID Wire Format` calls this out for off-chain tooling that may have integrated against the buggy form.
- **Event tags.** CIS-8004-native events use tags 240..247; CIS-2 inherited events use tags 251..255. Both ranges encode `token_id` in the standard `TokenIdU64` form post-redeploy.

## Render check

`source/index.rst` toctree updated to include `CIS/cis-8` and `CIS/cis-8004`. Both files use the same Sphinx directives (`.. list-table::`, `.. code-block:: rust`, `.. note::`, internal `:ref:` targets) as the existing CIS docs in this repo. Local doc8 lint not run; happy to fix anything CI flags.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+
+   * - Created
+     - May 21, 2026
+   * - Draft
+     - May 25, 2026
+   * - Supported versions
+     - | Smart contract version 1 (concordium-std v10+)
+   * - Standard identifier
+     - ``CIS-8``
+
```
tschudid, line 20: Does this need the following? 

   * - Requires
     - :ref:`CIS-0<CIS-0>`

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+The contract maintains only the current state of each registration. Full
+history is recoverable from the emitted events.
+
+Terminology and Conventions
+---------------------------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in :rfc:`2119`.
+
```
tschudid: In CIS-2 and CIS-3 this is in a different section. 
How much should the new standard fit the template from the old CIS?

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+    #[derive(Serialize, SchemaType)]
+    pub struct InitParams {
+        /// Single-account admin; MAY be transferred or renounced via
+        /// transferAdmin (see Admin section).
+        pub admin: AccountAddress,
+    }
+
+Initialises an empty ``registrations`` map and stores ``state.admin =
+Some(params.admin)``.
```
limemloh: How the smart contract is initialized is irrelevant for the spec and just an implementation detail

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+Terminology and Conventions
+---------------------------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in :rfc:`2119`.
+
+Rust types are expressed using ``concordium-std`` conventions (``Serialize``,
+``SchemaType``, ``Reject``). Schemas embedded in deployed modules are
+serialised in the Concordium binary schema format.
```
limemloh: This is not a problem as such, but it is not how we usually write these specs.
This is essentially telling people to understand some of the internals of the rust smart contract library in order to understand the serialization details of this spec. Usually we write out the actual serialization in the spec for each type.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+
+.. code-block:: rust
+
+    #[derive(Serial, DeserialWithState)]
+    #[concordium(state_parameter = "S")]
+    pub struct State<S: HasStateApi> {
+        /// Current registration keyed by ExternalKeyId.
+        pub registrations: StateMap<ExternalKeyId, Registration, S>,
+        /// Single-account admin model (see Admin section).
+        pub admin: Option<AccountAddress>,
```
limemloh: Having an admin seems like an implementation detail and is not relevant for interoperability which is the point of this spec.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+        pub owner:        AccountAddress,
+        pub external_key: ExternalKeyId,
+        pub proof_scheme: String,
+        pub metadata:     Vec<MetadataEntry>,
+        pub status:       RegistrationStatus,
+        pub last_updated: Timestamp,
+    }
+
+
+Contract State
```
limemloh: I don't think we should include the contract state in the spec, it is not important for interoperability

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+    #[concordium(state_parameter = "S")]
+    pub struct State<S: HasStateApi> {
+        /// Current registration keyed by ExternalKeyId.
+        pub registrations: StateMap<ExternalKeyId, Registration, S>,
+        /// Single-account admin model (see Admin section).
+        pub admin: Option<AccountAddress>,
+    }
+
+Only the current state is stored. Historical transitions MUST be recoverable
+from emitted events.
```
limemloh: Lets not dictate whether they can store historical information, but just emphasize the fact only the current state is needed.
```suggestion
By design only the current state needs to be stored. Historical transitions MUST be recoverable
from emitted events.
```

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+Registration
+^^^^^^^^^^^^
+
+.. code-block:: rust
+
+    #[derive(Serialize, SchemaType, Clone, Debug)]
+    pub struct Registration {
+        pub owner:        AccountAddress,
+        pub external_key: ExternalKeyId,
+        pub proof_scheme: String,
```
tschudid: Why is this not of the type Proof?

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+   * - :ref:`CIS-8-ExternalKeyRegistered`
+     - 250
+   * - :ref:`CIS-8-ExternalKeyRevoked`
+     - 251
+   * - :ref:`CIS-8-UpdateMetadata`
+     - 252
+   * - :ref:`CIS-8-Upgraded`
+     - 253
+   * - :ref:`CIS-8-AdminTransferred`
+     - 254
```
limemloh: These tags are clashing with events in CIS-2, so it would not be possible to have a contract implementing both CIS-8 and CIS-2 which might be a problem.
At least easy to fix until launch.
Using 235, 234, 233, 232, 231 instead would make it compatible with every other CIS.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+   * - Event
+     - Tag
+   * - :ref:`CIS-8-ExternalKeyRegistered`
+     - 250
+   * - :ref:`CIS-8-ExternalKeyRevoked`
+     - 251
+   * - :ref:`CIS-8-UpdateMetadata`
+     - 252
+   * - :ref:`CIS-8-Upgraded`
+     - 253
```
limemloh: No need for upgrade event, the chain already pu a trace element representing an upgrade on chain.

source/CIS/cis-8004.rst
```diff
@@ -0,0 +1,1143 @@
...
+
+.. note::
+
+   Earlier draft deployments of this contract (pre 2026-05-25) incorrectly
+   typed the CIS-8004-native entrypoints as bare ``u64`` LE (8 bytes,
+   no length prefix). That deviation was a contract bug, not a spec
+   decision; the redeployed reference instances listed in
+   :ref:`CIS-8004-DeploymentStatus` use the spec-correct ``TokenIdU64``
+   form throughout. Off-chain tooling that talked to the pre-redeploy
+   instances will need updating.
```
limemloh: Should this be included?

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+
+
+Canonical Signed Message
+------------------------
+
+.. _CIS-8-CanonicalMessage:
+
+CIS-8 verifies that the holder of the external private key has authorised the
+binding to a specific Concordium account on a specific contract on a specific
+network. The contract reconstructs the canonical message; it MUST NEVER accept
```
tschudid, line 231: We should clearly define how the message is formed (not just using the `self.serial()` from below. 

Important: We need to take care about ambiguous parsing. In particular, any element that does not have a fixed length should have its length prepended.

tschudid: See how ETH signatures are constructed.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+solana-ed25519
+^^^^^^^^^^^^^^
+
+| Algorithm: Ed25519
+| Public key encoding: ``ed25519`` (32 bytes)
+| Signature format: 64 bytes ``R || S``
+
+Verification:
+
+1. Use ``message`` directly as the signed payload (no additional wrapping).
```
tschudid, line 312: Add a reference to this envelope.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+.. code-block:: rust
+
+    #[derive(Serialize, SchemaType)]
+    pub struct RegisterParams {
+        pub external_key: ExternalKeyId,
+        pub proof:        Proof,
+        pub metadata:     Vec<MetadataEntry>,
+    }
+
+Authorisation
```
tschudid: The rendering of this gives it a larger font size than `register`.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+    }
+
+Initialises an empty ``registrations`` map and stores ``state.admin =
+Some(params.admin)``.
+
+.. _CIS-8-register:
+
+register
+^^^^^^^^
+
```
tschudid: There should be a single sentence giving a high level explanation of this interface.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,765 @@
...
+.. list-table::
+   :header-rows: 1
+
+   * - Code
+     - Name
+     - Meaning
+   * - .. _CIS-8-InvalidProof:
+
+       -7100
```
tschudid: Why do we start at `7100`?

source/CIS/cis-8.rst
andynelson: completely?

source/CIS/cis-8004.rst
```diff
@@ -0,0 +1,1133 @@
...
+   view entrypoints so integrators can sanity-check the currently-configured
+   dependency addresses; see :ref:`CIS-8004-DeploymentStatus`.
+
+
+Entrypoints
+-----------
+
+.. _CIS-8004-init:
+
+init
```
limemloh: As with CIS-8, this is just an implementation detail and should not be part of the spec.

source/CIS/cis-8.rst
MarirosansP: Wanted to clarify before changing the data shape. Two reads of the comment:

**(a)** Store the full `Proof` (scheme + signature bytes) on the on-chain `Registration` record. That gives consumers the ability to independently re-verify a registration after the fact — at the cost of ~64 bytes of permanent state per registration (the signature isn't otherwise needed once `register` has completed). Currently we only retain the scheme name since the signature is consumed at register time.

**(b)** Keep just the scheme name but make it a typed enum (`ProofScheme`) instead of a plain string — same on-chain footprint, more type-safe, but locks the set of allowed schemes at the spec level.

I've added a sentence to the `Registration` description in the latest commit explaining the current "scheme only, not signature" choice and pointing readers at the `ExternalKeyRegistered` event payload if they want to retain the signature off-chain. Happy to switch to either (a) or (b) — let me know which you had in mind.

limemloh (approved):

tschudid (approved):
Looks good to me for a first version. 
Need to check the signature proof in more details for the next version.

source/CIS/cis-8.rst
```diff
@@ -0,0 +1,723 @@
...
+     - Notes
+   * - Concordium DevNet P-11
+     - ``<23, 0>``
+     - Concordium-owned redeploy 2026-05-25
+   * - Concordium TestNet
+     - ``<12794, 0>``
+     - Concordium-owned redeploy 2026-05-25; same module bytes as DevNet
+
+
+Security Considerations
```
tpryds, line 682: One could add in this section that adding "concordium_account" to the canonical messge ensure that the proof is tide to a registration by this concordium account.

source/CIS/cis-8004.rst
```diff
@@ -0,0 +1,1133 @@
...
+Abstract
+========
+
+A standard interface for an Agent Registry contract that records the identity
+of autonomous software agents as transferable CIS-2 NFTs.
+
+CIS-8004 is the Concordium implementation of the Identity Registry component
+of `ERC-8004 v2 (Trustless Agents) <https://eips.ethereum.org/EIPS/eip-8004>`_.
+The Reputation Registry and Validation Registry components of ERC-8004 v2 are
+out of scope for this version.
```
tpryds, line 32: Suggest to delete "for this version". Hopefully we will not have more versions of the standard.

source/CIS/cis-8004.rst
```diff
@@ -0,0 +1,1133 @@
...
+
+Each agent is identified on chain by a 64-bit token id. The canonical
```
tpryds, line 34: Did you consider using CAIP-19 here?

source/CIS/cis-8004.rst
```diff
@@ -0,0 +1,1133 @@
...
+        pub registered_at:      Timestamp,
+        pub revoked_at:         Option<Timestamp>,
+        pub revocation_reason:  Option<String>,
+        pub on_chain_metadata:  StateMap<String, Vec<u8>, S>,
+    }
+
+The ``owner_account`` is **not stored** on the record; it is derived from the
+CIS-2 ownership state (``cis2_state.token_owner(token_id)``) at query time.
+
+The ``agent_wallet`` **is** stored. It has an independent lifecycle from
```
tpryds, line 228: It is inconsistent to use the word "wallet" rather than "account"

PR #86, opened 2026-05-28
Add cis 8 and cis 8004

limemloh:
## Purpose

Based on #84, fixes the build and removes CIS-10 for now


limemloh:
Closed in favor of merging #84 directly

PR #88, opened 2026-05-28
Revamp cis8 and cis8004

limemloh:
## Purpose

Update CIS-8 and CIS-8004 to follow the style of other CIS standards


source/CIS/cis-8004.rst
```diff
@@ -1,942 +1,818 @@
...
+a :ref:`CIS-8004-Timestamp` (``deadline``),
+a :ref:`CIS-8004-ContractAddress` (``contract_address``),
+and a :ref:`CIS-8004-BlockHash` (``genesis_hash``)::
 
+  SetAgentWalletSigningData ::= ("CIS-8004/v1/setAgentWallet": Byte²⁶)
+                                 (token_id: AgentTokenId)
+                                 (new_wallet: AccountAddress)
+                                 (deadline: Timestamp)
+                                 (contract_address: ContractAddress)
+                                 (genesis_hash: BlockHash)
```
soerenbf, line 251: I think @tpryds had a comment on the original PR related to this, but should we consider _not_ using wallet as a term? It seems incorrect as we're talking about accounts here.

source/CIS/cis-8004.rst
```diff
@@ -1,942 +1,818 @@
...
-        pub token_id: AgentTokenId,
-        pub key:      String,
-        pub value:    Vec<u8>,
-    }
+``supports``
+^^^^^^^^^^^^
 
-Authorisation
-    Sender MUST equal the current CIS-2 owner.
+As specified in :ref:`CIS-0<CIS-0>`.
```
soerenbf: should we reference cis-0 the same way we do with cis-2 here, i.e. move it to the previous paragraph, and maybe order them by the cis index?

soerenbf (approved):
Nice job!

I think it'd be nice if we could get rid of the "wallet" term altogether and use "account" instead. I know this is not your call, but I'm raising it here so we can at least come to an agreement and perhaps make a record of the decision to change it? 🤷

source/CIS/cis-8004.rst
limemloh: I agree on the point, but I also know there is a desire to stay as close to the [ERC-8004](https://eips.ethereum.org/EIPS/eip-8004) as possible, and they use wallet here.

soerenbf: oh, alright i see the point then.

